PR #7, opened 2015-12-14
use -a to define plugin

timogoebel:
This commit changes the call to letsencrypt-auto to use the parameter -a
to choose the plugin used to deploy the certificates.

To quote an example from the beta invite:
./letsencrypt-auto certonly -a standalone \
  -d example.com -d www.example.com


danzilio:
Thank you @timogoebel!

PR #8, opened 2015-12-30
add optional cron job for certificate renewal

hdeadman:
This creates a cronjob that sets up a daily job to re-run letsencrypt-auto with same arguments plus the --keep-until-expiring argument so that certificate will be renewed when it is close to expiring. I tested this on a server with two websites on it and it scheduled two jobs which ran OK but didn't renew certificate because they aren't close to expiring. When I ran the command manually with the renew-by-default argument it did renew the certificate (but early).

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -14,12 +14,16 @@
 # [*additional_args*]
 #   An array of additional command line arguments to pass to the
 #   `letsencrypt-auto` command.
+# [*manage_cron*]
+#   Boolean indicating whether or not to schedule cron job for renewal, 
+#   runs daily but only renews if near expiration e.g within 10 days. 
 #
 define letsencrypt::certonly (
   Array[String]                           $domains          = [$title],
   Enum['apache', 'standalone', 'webroot'] $plugin           = 'standalone',
   String                                  $letsencrypt_path = $letsencrypt::path,
   Optional[Array[String]]                 $additional_args  = undef,
+  Boolean                                 $manage_cron      = $letsencrypt::manage_cron,
 ) {
 
   $command = inline_template('<%= @letsencrypt_path %>/letsencrypt-auto certonly -a <%= @plugin %> -d <%= @domains.join(" -d ")%><% if @additional_args %> <%= @additional_args.join(" ") %><%end%>')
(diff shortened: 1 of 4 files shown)


manifests/certonly.pp
```diff
@@ -31,4 +35,16 @@
...
+  
+  if ($manage_cron) {
+    $renewcommand = inline_template('<%= @letsencrypt_path %>/letsencrypt-auto certonly -a <%= @plugin %> --keep-until-expiring -d <%= @domains.join(" -d ")%><% if @additional_args %> <%= @additional_args.join(" ") %><%end%>')
+    $cron_hour = fqdn_rand (24, $title) # 0 - 23, seed is title plus fqdn
+    $cron_minute = fqdn_rand( 60, $title ) # 0 - 59, seed is title plus fqdn
+    cron { "letsencrypt renew cron ${title}":
+      command => $renewcommand,
+      user    => root,
+      hour    => $cron_hour,
+      minute  => $cron_minute,
```
danzilio: The syntax check is failing due to trailing whitespace on this line.

manifests/certonly.pp
```diff
@@ -31,4 +35,16 @@
     creates => $live_path,
     require => Class['letsencrypt'],
   }
+  
+  if ($manage_cron) {
```
danzilio: Can you remove the parentheses here? I'd like to keep the style consistent.

manifests/certonly.pp
```diff
@@ -31,4 +35,16 @@
...
+    $renewcommand = inline_template('<%= @letsencrypt_path %>/letsencrypt-auto certonly -a <%= @plugin %> --keep-until-expiring -d <%= @domains.join(" -d ")%><% if @additional_args %> <%= @additional_args.join(" ") %><%end%>')
+    $cron_hour = fqdn_rand (24, $title) # 0 - 23, seed is title plus fqdn
```
danzilio: Can you remove the space between the function name and the parentheses here?

manifests/certonly.pp
```diff
@@ -31,4 +35,16 @@
...
+    $cron_minute = fqdn_rand( 60, $title ) # 0 - 59, seed is title plus fqdn
```
danzilio: Can you also remove the spaces between the parentheses and the arguments here?

manifests/certonly.pp
```diff
@@ -14,12 +14,16 @@
...
+# [*manage_cron*]
+#   Boolean indicating whether or not to schedule cron job for renewal, 
+#   runs daily but only renews if near expiration e.g within 10 days. 
 #
 define letsencrypt::certonly (
   Array[String]                           $domains          = [$title],
   Enum['apache', 'standalone', 'webroot'] $plugin           = 'standalone',
   String                                  $letsencrypt_path = $letsencrypt::path,
   Optional[Array[String]]                 $additional_args  = undef,
+  Boolean                                 $manage_cron      = $letsencrypt::manage_cron,
```
danzilio: This seems to be a per-cert setting, let's move the default value into this class and remove the `$manage_cron` parameter from `init.pp` and it's default from `params.pp`.

danzilio:
Hi @hdeadman! Thanks for the PR! I've left some comments inline. Also, can you please add a test for this feature?

hdeadman:
OK, thanks for the feedback. I made the updates you requested.

danzilio:
Thanks!

PR #9, opened 2016-01-03
Add --agree-tos to commands

stephenwade:
Without this switch, the command fails with the error "Unable to register an account with ACME server".


danzilio:
Thanks!!!

PR #10, opened 2016-01-04
Added webroot_paths parameter

stephenwade:
This lets the plugin work correctly with `plugin => 'webroot'` by supplying `--webroot-path` parameters for each domain.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -9,24 +9,36 @@
 #   Namevar. An array of domains to include in the CSR.
 # [*plugin*]
 #   The authenticator plugin to use when requesting the certificate.
+# [*webroot_paths*]
+#   An array of webroot paths for the domains in `domains`.
+#   Required if using `plugin => 'webroot'`. If `domains` and
+#   `webroot_paths` are not the same length, `webroot_paths`
+#   will cycle to make up the difference.
 # [*letsencrypt_path*]
 #   The path to the letsencrypt installation.
 # [*additional_args*]
 #   An array of additional command line arguments to pass to the
 #   `letsencrypt-auto` command.
 # [*manage_cron*]
-#   Boolean indicating whether or not to schedule cron job for renewal, 
-#   runs daily but only renews if near expiration e.g within 10 days. 
+#   Boolean indicating whether or not to schedule cron job for renewal.
+#   Runs daily but only renews if near expiration, e.g. within 10 days. 
 #
 define letsencrypt::certonly (
   Array[String]                           $domains          = [$title],
   Enum['apache', 'standalone', 'webroot'] $plugin           = 'standalone',
+  Optional[Array[String]]                 $webroot_paths    = undef,
   String                                  $letsencrypt_path = $letsencrypt::path,
   Optional[Array[String]]                 $additional_args  = undef,
   Boolean                                 $manage_cron      = false,
 ) {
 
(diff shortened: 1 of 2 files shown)


hdeadman:
I am currently doing this:
```
		letsencrypt::certonly { $sitewww:
		  domains         => [$sitedomain, $sitewww],
		  plugin          => 'webroot',
		  additional_args => ["--webroot-path /srv/wordpress/$sitename"],
		  require         => Wp::Site[$sitename],
		  manage_cron     => true,
		}
```

I have multiple domains but they are actually just multiple names for the cert in the top level domain and the www sub-domain. The number of domains and the number of webroot arguments don't necessarily have to match up. In my case would I list the same webroot twice if I were using your new argument?

Since the certificates are free I am just generating a separate certificate for each top level domain and putting multiple names in the certificates for the sub-domains. 

Wouldn't the cron command have to be updated as well?

stephenwade:
You're right, I forgot about updating the cron command.

I didn't think about the use case of multiple domains with one webroot. I'll update this when I get home to support one webroot path for multiple domains, and to fix the cron command.

danzilio:
@stephenwade if you update the cron command, i'll merge this

stephenwade:
Sorry about the delay.

In addition to fixing the cron command, I took @hdeadman's idea and made the `webroot_paths` array cycle. That way, you can use only one path and have it applied to all the domains.

danzilio:
Thank you! One last thing, could you squash your commits? I'd like to keep the commit history nice and clean.

Once this is merged I'll push a new release to the Forge.

stephenwade:
ta-dah!

danzilio:
Beautiful! Thanks for working on this :)

hdeadman:
Does this support multiple webroots each with own one or more domains? I don't need that feature but I think it is supported by letsencrypt according to this thread; 

https://community.letsencrypt.org/t/multiple-domains-webroot-paths-using-webroot-plugin/7982
> You can add up to 100 different (sub)domains in one certificate. It would be exceptional if all those domains would fall under one webroot

stephenwade:
I can’t think of a good syntax for multiple webroots, each with multiple domains, in the same resource. If someone needed that, I think the best way would be either to use multiple resources, each with one webroot and its domains, or to copy the same webroot multiple times in the list.

So, either
```
letsencrypt::certonly { 'foo':
  plugin        => 'webroot',
  domains       => [ 'a.foo.example.com', 'b.foo.example.com' ],
  webroot_paths => [ '/www/foo' ],
}

letsencrypt::certonly { 'bar':
  plugin        => 'webroot',
  domains       => [ 'a.bar.example.com', 'b.bar.example.com' ],
  webroot_paths => [ '/www/bar' ],
}
```

or
```
letsencrypt::certonly { 'foobar':
  plugin        => 'webroot',
  domains       => [ 'a.foo.example.com', 'b.foo.example.com', 'a.bar.example.com', 'b.bar.example.com' ],
  webroot_paths => [ '/www/foo', '/www/foo', '/www/bar', '/www/bar' ],
}
```

stephenwade:
On the command line, you could do

`--webroot-path /www/foo -d a.foo.example.com -d b.foo.example.com --webroot-path /www/bar -d a.bar.example.com -d b.bar.example.com`

but I'm not sure how to express that idea with the Puppet syntax. The only option I've thought of is creating a separate resource for the `webroot` plugin so that webroots could have domains as a sub-array or something.

stephenwade:
Hey @danzilio, just wondering when this is going to make it to Puppet Forge.

danzilio:
@stephenwade sorry about the delay on this! I promise I'll get a new release to the Forge today!

danzilio:
@stephenwade just pushed `v0.4.0` to the Forge!

stephenwade:
Awesome!

PR #11, opened 2016-01-31
Install LE package on EL7, Ubuntu 16.04 and Debian 9+

domcleal:
A new $install_method parameter to letsencrypt allows selection between
a git/VCS based installation and a package based installation.  It
defaults to packages on newer OSes where they're available - in EPEL7,
and upcoming Debian and Ubuntu releases.

Installation has been moved into letsencrypt::install and the LE command
itself is determined by the installation method.

### DIFF
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,5 +1,8 @@
 fixtures:
   repositories:
+    epel:
+      repo: 'https://github.com/stahnma/puppet-module-epel.git'
+      ref: '1.2.2'
     inifile:
       repo: 'https://github.com/puppetlabs/puppetlabs-inifile.git'
       ref: '1.4.2'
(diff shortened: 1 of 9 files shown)


danzilio:
Thanks @domcleal, this is awesome!! Do you think we should manage EPEL in this? I'm just wondering if `$install_method = 'package'` is a sane default for EL if we're not expressing the dependency on EPEL. Also, what do you think about adding a default case for `$::osfamily` in `letsencrypt::params` with `$install_method = 'vcs'`?

domcleal:
Yeah, I share your concern about it not working out of the box on EL7, but I'm hesitant to manage EPEL as it could be quite an intrusive change for some users.

I updated the PR to add a soft dependency on the common [stahnma/epel](https://forge.puppetlabs.com/stahnma/epel) module, with correct ordering of the Yumrepo vs. Package resource, but disabled by default (so as not to require it in metadata.json).

I've also dropped the unsupported OS case and defaulted it to a VCS installation as you suggested.

danzilio:
This is excellent! Thank you @domcleal!

PR #12, opened 2016-02-03
Change Puppet 4 specifics to support 3.4+

domcleal:
Future/Puppet 4 parser specifics have been removed to support Puppet
3.x, though requiring a minimum of 3.4 for the contain() function.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -40,15 +40,7 @@ class letsencrypt::config (
     }
   }
 
-  $_config.each |$setting, $value| {
-    ini_setting { "${config_file} ${setting} ${value}":
-      ensure  => present,
-      path    => $config_file,
-      section => '',
-      setting => $setting,
-      value   => $value,
-      require => File['/etc/letsencrypt'],
-    }
-  }
+  $_config_joined = join_keys_to_values($_config, '=')
+  letsencrypt::config::ini { $_config_joined: }
 
 }
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -36,20 +36,27 @@
 #   A flag to allow using the 'register-unsafely-without-email' flag.
 #
 class letsencrypt (
-  Optional[String]       $email               = undef,
-  String                 $path                = $letsencrypt::params::path,
-  String                 $repo                = $letsencrypt::params::repo,
-  String                 $version             = $letsencrypt::params::version,
-  String                 $config_file         = $letsencrypt::params::config_file,
-  Hash[String, Any]      $config              = $letsencrypt::params::config,
-  Boolean                $manage_config       = $letsencrypt::params::manage_config,
-  Boolean                $manage_install      = $letsencrypt::params::manage_install,
-  Boolean                $manage_dependencies = $letsencrypt::params::manage_dependencies,
-  Boolean                $configure_epel      = $letsencrypt::params::configure_epel,
-  Enum['package', 'vcs'] $install_method      = $letsencrypt::install_method,
-  Boolean                $agree_tos           = $letsencrypt::params::agree_tos,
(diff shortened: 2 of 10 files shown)


danzilio:
Thanks for this @domcleal! I mentioned this to some folks in #voxpupuli on Freenode today, here are some of the thoughts I fleshed out. I originally set out to make this module compatible with Puppet > 4 because I didn't want to maintain Puppet 3 compatibility and wanted to set an example for Puppet 4 modules going forward. Since releasing the module I've had a few people ask me to add backwards compatibility. I'm starting to think that my position on backwards compatibility, while principled, may not be practical. I wonder how many people are stuck on Puppet 3. I'm definitely interested in making this module as helpful to as many people as possible and that may mean supporting Puppet 3 for some period of time.

How long do you think we should maintain Puppet 3 compatibility, should we go forward with this?

manifests/config.pp
```diff
@@ -40,15 +40,7 @@
...
-      ensure  => present,
-      path    => $config_file,
-      section => '',
-      setting => $setting,
-      value   => $value,
-      require => File['/etc/letsencrypt'],
-    }
-  }
+  $_config_joined = join_keys_to_values($_config, '=')
+  letsencrypt::config::ini { $_config_joined: }
```
danzilio: We could probably use [`create_ini_settings`](https://github.com/puppetlabs/puppetlabs-inifile/blob/master/lib/puppet/parser/functions/create_ini_settings.rb) for this.

manifests/init.pp
```diff
@@ -36,20 +36,27 @@
...
+  $path                = $letsencrypt::params::path,
+  $repo                = $letsencrypt::params::repo,
+  $version             = $letsencrypt::params::version,
+  $config_file         = $letsencrypt::params::config_file,
+  $config              = $letsencrypt::params::config,
+  $manage_config       = $letsencrypt::params::manage_config,
+  $manage_install      = $letsencrypt::params::manage_install,
+  $manage_dependencies = $letsencrypt::params::manage_dependencies,
+  $configure_epel      = $letsencrypt::params::configure_epel,
+  $install_method      = $letsencrypt::install_method,
```
danzilio: This param references itself! Should be:

```puppet
$install_method      = $letsencrypt::params::install_method
```

manifests/config.pp
domcleal: This seems to fail on Puppet 4.0 and 4.1, but not on 4.2 or 3.x, with a duplicate declaration error:

    Evaluation Error: Error while evaluating a Function Call, Duplicate declaration: Ini_setting[[] email] is already declared; cannot redeclare at /home/travis/build/domcleal/puppet-letsencrypt/spec/fixtures/modules/letsencrypt/manifests/config.pp:48:3

I can't see any reason for it, and removing the create_resources() call inside the function shows that the ini_settings resource isn't there from somewhere else.  I suppose it's a Puppet bug, but not one I'm aware of.

manifests/init.pp
domcleal: Thanks, fixed.

domcleal:
> How long do you think we should maintain Puppet 3 compatibility, should we go forward with this?

I'd want to use it for at least a year I think.  I'm hoping I may be able to add it to Foreman's installer modules, and that whole stack supports 3.x still (and only recent 4.x).

manifests/config.pp
domcleal: As discussed on IRC, this was PUP-4709.  I sent inifile a PR at https://github.com/puppetlabs/puppetlabs-inifile/pull/195 to remove the square brackets, but for now I've restored the letsencrypt::config::ini define.

danzilio:
Thanks for your work on this @domcleal! Once the build goes green, I'll merge this, write up the changelog, and release as 1.0! CC: @pgassmann

PR #13, opened 2016-02-04
updating changelog and prepping metadata for release as 1.0

danzilio:


CHANGELOG.md
```diff
@@ -2,6 +2,14 @@
 All notable changes to this project will be documented in this file. This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased][unreleased]
+## Added
+- Backwards compatibility with Puppet >= 3.4
+- Ability to select the `letsencrypt` install method using the `install_method` parameter. Current supported options are `package` and `vcs`.
+- The `manage_install` parameter now lets the user select whether they want to manage the installation of `letsencrypt` with this module.
+- The `configure_epel` parameter now lets the user manage the EPEL repository on EL systems.
```
domcleal: Perhaps note that the epel module's needed for this to work.

domcleal:
Worth noting either under `install_method` or `Breaking` that the default behaviour's changed on some OSes to install the package rather than from source.

danzilio:
Hey @domcleal, sorry I've let this sit for so long! I just updated `CHANGELOG` but I also decided to change the default `install_method` for `RedHat` back to `vcs`. I just think that setting `install_method` to `package` for all of `RedHat` is overly broad when the `letsencrypt` package is not available in core `RedHat` repositories yet. I'm torn between leaving the default as `vcs` or changing the default for `manage_epel` to `true` to accommodate. I think the least intrusive means is to just leave the default of `vcs`. Thoughts?

domcleal:
Yeah, I understand your reasoning and desire to keep it working out of the box, I can't argue with that.

Being a chronic software packager I dislike the `vcs` installation, so I'd prefer to force the user to choose between using a VCS installation or configuring EPEL (by setting `manage_epel` or using another module) when installing the module.

I can't help but think the majority of users would prefer to enable EPEL over using a VCS installation, so setting `manage_epel` to true by default sounds reasonable.

danzilio:
Yeah I'm not particularly a fan of the `vcs` method, it was mostly a stopgap until packaging became prevalent. Now that you mention it, I think you're right that the majority would prefer to configure EPEL over the VCS install...

Setting `manage_epel` to `true` means we should make `epel` a hard dependency in `metadata.json`...

domcleal:
> Setting manage_epel to true means we should make epel a hard dependency in metadata.json...

Yes, quite probably - if it's true by default then it should probably always be installed.

danzilio:
OK! Just pushed those changes!

domcleal:
I think that's OK, thanks @danzilio.  :+1:

danzilio:
Ok! I'll let the build finish and then push the release! This is a big change so I'm also going to send an email out to the puppet-users mailing list announcing the release.

Issue #14, opened 2016-02-15
Update to the latest lets-encrypt version: v0.4.0


Kriechi:
Is there a reason or incompatibility why the version currently defaults to `v0.1.0` [here](https://github.com/danzilio/puppet-letsencrypt/blob/9bcf3c660dd9c69e7fcabbbb95fec74637fa70bd/manifests/params.pp#L11)?
lets-encrypt released `v0.4.0` a couple of days ago, and I think I would be nice to upgrade the default.

And with this new release there is now support for a `renew` command, which could be used in the cron job.
https://letsencrypt.readthedocs.org/en/latest/using.html#renewal

danzilio:
No deliberate reason for this. PRs are welcome ;)

danzilio:
Closed via #15

PR #15, opened 2016-02-22
Changing default version to 0.4.0

as0bu:
This should close issue #14

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -9,7 +9,7 @@ class letsencrypt::params {
   $config_file         = '/etc/letsencrypt/cli.ini'
   $path                = '/opt/letsencrypt'
   $repo                = 'git://github.com/letsencrypt/letsencrypt.git'
-  $version             = 'v0.1.0'
+  $version             = 'v0.4.0'
   $config              = {
     'server' => 'https://acme-v01.api.letsencrypt.org/directory',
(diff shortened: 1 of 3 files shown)


danzilio:
:+1:

Kriechi:
Thanks!

No usage of the [renew command](https://letsencrypt.readthedocs.org/en/latest/using.html#renewal) though... [see here](https://github.com/danzilio/puppet-letsencrypt/blob/master/manifests/certonly.pp#L62)

as0bu:
@Kriechi It looks like the current command will renew it, although on a per certificate basis. From my understanding, and please correct me if I'm wrong, the current cronjob will renew, but renews each certificate rather than all at the same time. I'm assuming that if all of your letsencrypt certificates are managed via this module then the end result is the same. Is that not the case?

>Running letsencrypt renew will renew all certificates that are within the renewal window (by default, certificates within 30 days of expiry). If you would like to renew a single certificate, you should call letsencrypt certonly -d with the exact same domains that are specified in the certificate you are trying to renew. For example:

>letsencrypt certonly --keep-until-expiring --webroot -w /var/www/example.com -d example.com,www.example.com -w /var/www/thing -d thing.is,m.thing.is

This module essentially does the "single certificate" renewal.

Issue #16, opened 2016-02-25
vcs-installed letsencrypt-auto creates '~' directory


mheistermann:
When you install letsencrypt using the VCS method, it will create a virtualenv, which would normally be located in ~/.local/share/letsencrypt:

Excerpt from letsencrypt-auto:
```
XDG_DATA_HOME=${XDG_DATA_HOME:-~/.local/share}
VENV_NAME="letsencrypt"
VENV_PATH=${VENV_PATH:-"$XDG_DATA_HOME/$VENV_NAME"}
```
However, puppet exec's will not set $HOME, which will stop Debian's /bin/sh, dash, from expanding ~, so it creates a literal '~' folder.

I'm currently fixing this, will create a pull request soon.

PR #18, opened 2016-02-25
Feature: Allow running a command after successful cronjob.

mheistermann:
Using the `cron_success_cmd` parameter for `certonly`, you can
run commands when the renewal command was successful, e.g.
reload the webserver to use the renewed certs.

Not sure about letsencrypt return values, so the cmd might be
executed even if the cert is unchanged, doesn't matter for this
usecase though.

(This PR includes the bugfix for #16 for easy fast forward merging, assuming you want to merge the bugfix)

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -51,11 +51,13 @@ define letsencrypt::certonly (
   $command = "${command_start}${command_domains}${command_end}"
   $live_path = inline_template('/etc/letsencrypt/live/<%= @domains.first %>/cert.pem')
 
+  $venv_path_var = "VENV_PATH=${letsencrypt::venv_path}"
   exec { "letsencrypt certonly ${title}":
-    command => $command,
-    path    => $::path,
-    creates => $live_path,
-    require => Class['letsencrypt'],
+    command     => $command,
+    path        => $::path,
+    environment => [$venv_path_var],
+    creates     => $live_path,
+    require     => Class['letsencrypt'],
   }
 
   if $manage_cron {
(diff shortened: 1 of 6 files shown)


danzilio:
This one too!

manifests/certonly.pp
```diff
@@ -30,6 +30,7 @@
   $letsencrypt_command = $letsencrypt::command,
   $additional_args     = undef,
   $manage_cron         = false,
+  $cron_success_cmd    = undef,
```
danzilio: For consistency's sake, could we rename this param to `cron_success_command`?

danzilio:
@mheistermann this looks good. See my note on the parameter name. Could you please add a test for this?

mheistermann:
I renamed the parameter and added a test, thanks for the suggestion, had forgotten about it.

mheistermann:
Added some documentation for manage_cron and the newly introduced cron_success_command parameter to README and the latter to the CHANGELOG.

danzilio:
@mheistermann looks like this needs a rebase!

danzilio:
@mheistermann ping

tomgillett:
@mheistermann this would be super-helpful functionality!

danzilio:
@tomgillett @mheistermann I should be able to rebase this myself in a bit

tomgillett:
That would be great, thanks @danzilio :)

<off topic> Might you be interested in a PR for "environment" support as well? I need to run the client behind an outbound proxy. This means that HTTP_PROXY and HTTPS_PROXY need to be set. Currently I've added a new parameter that I'm concat-ing with the VENV_PATH variable.</off topic>

danzilio:
@tomgillett absolutely! Send it along!

danzilio:
Changes merged in #32

PR #23, opened 2016-03-29
changed rsa key size from 2048 (standard) to 4096

fpieters:
The title says it all. :+1:

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -12,7 +12,8 @@ class letsencrypt::params {
   $repo                = 'git://github.com/letsencrypt/letsencrypt.git'
   $version             = 'v0.4.0'
   $config              = {
-    'server' => 'https://acme-v01.api.letsencrypt.org/directory',
+    'server'           => 'https://acme-v01.api.letsencrypt.org/directory',
+    'rsa-key-size'     => '4096',
   }
 
   if $::operatingsystem == 'Debian' and versioncmp($::operatingsystemrelease, '9') >= 0 {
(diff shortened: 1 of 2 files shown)


fpieters:
Apparently there are problems with Red Hat and CentOS. :(

danzilio:
hey @fpieters! I've thought about this a bit and I think I'd like to leave this up to the user instead of forcing a default on them.

Kriechi:
:+1: Better and more secure defaults are always a good idea!

danzilio:
Definitely. What are the implications of this change for existing users when they upgrade the module? Will this result in new signing requests if the user had accepted the previous defaults? I just want to understand the impact on our users before making a change like this.

pgassmann:
I would leave it to the upstream default. Longer Keys lead to more cpu and
power usage for questionably more security.

2048 is safe enough for now. For these short-lived certificates, it could
be updated when recommended by experts or when letsencrypt updates their
default.

See also the first Google result:
https://www.yubico.com/2015/02/big-debate-2048-4096-yubicos-stand/

Or
http://security.stackexchange.com/questions/65174/4096-bit-rsa-encryption-keys-vs-2048

>
> Advisories recommend 2048 for now. Security experts are projecting that
> 2048 bits will be sufficient for commercial use until around the year 2030.


2016-04-04 16:32 GMT+02:00 David Danzilio <notifications@github.com>:

fpieters:
Leaving the key size up to the user and/or the letsencrypt default is fine by me. But personally I do use the increased size, since I think its worth the little cpu increase.

danzilio:
Oh I definitely agree with the rationale behind this change, I just want to make sure our changes are predictable and frictionless for our users. Since we don't explicitly set a key size in this module, I'm going to say we should let the client determine the sanity of its own defaults. I appreciate the contribution though!

Issue #26, opened 2016-03-31
Command Questions


lazyfrosch:
During implementation yesterday I came up to a few "problems":

* The cron outputs senseless info every time:
```
Checking for new version...
Requesting root privileges to run letsencrypt...
   /opt/letsencrypt/.venv/bin/letsencrypt --agree-tos certonly -a webroot --keep-until-expiring --webroot-path /var/www/html -d test.example.com
```
* Should we use `--no-self-upgrade` on cron and certonly?
* Should we use `/opt/letsencrypt/.venv/bin/letsencrypt`, and call letsencrypt-auto only once?

danzilio:
@lazyfrosch merged your PR for this. Thanks!

PR #27, opened 2016-04-02
Split letsencrypt commands, so -auto is only used once

lazyfrosch:
So we avoid fiddling with the venv every time, and the cron is now
silent.

refs GH-26


danzilio:
thanks @lazyfrosch! i'm just trying to get a few more things settled with this before shipping a release to the forge. i'll do some testing today and hopefully get it out after that

lazyfrosch:
@danzilio take your time :wink:

PR #34, opened 2016-05-23
Add a way to change package name and command

glorpen:
This change adds package_name and package_command params
and as a bonus support for Gentoo. Test for this use case
was also created.

Name and command params default to their old values,
so change should be fully backward compatible.

### DIFF
--- a/manifests/install.pp
+++ b/manifests/install.pp
@@ -24,12 +24,16 @@
 # [*package_ensure*]
 #   The value passed to `ensure` when installing the client with the `package`
 #   method.
+# [*package_name*]
+#   Name of package to use when installing the client with the `package`
+#   method. 
 #
 class letsencrypt::install (
   $manage_install      = $letsencrypt::manage_install,
   $manage_dependencies = $letsencrypt::manage_dependencies,
   $configure_epel      = $letsencrypt::configure_epel,
   $install_method      = $letsencrypt::install_method,
+  $package_name        = $letsencrypt::package_name,
   $package_ensure      = $letsencrypt::package_ensure,
   $path                = $letsencrypt::path,
   $repo                = $letsencrypt::repo,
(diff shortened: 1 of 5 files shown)


manifests/install.pp
```diff
@@ -53,13 +57,13 @@
       revision => $version,
     }
   } else {
-    package { 'letsencrypt':
+    package { $package_name:
```
domcleal: Nitpick, I'd suggest to keep the resource title the same (e.g. `Package[letsencrypt]`) but supply the `name` parameter with the varying value, then it's less confusing if `$package_name` isn't set correctly or somebody wants to add a relationship on the resource.

domcleal:
Could you also update letsencrypt::certonly to use the command name?

The package has also changed to certbot in EPEL7, but I'd be happy to send a followup PR to rename the package there too.

Note that tests haven't run, perhaps they're not enabled correctly on this repo?

danzilio:
Hmm not sure why the tests haven't run on this...

manifests/install.pp
glorpen: In the adding relationship case, wouldn't it be expected to have working `Package[$my_pkg]` when providing `$package_name`, eg. `class {'letsencrypt': package_name => $my_pkg}`? After all it is named "package_name".

I'm not really sure about the "right way" here ... *(reads puppet documentation)* ... I guess I don't find the aliasing feature as clean ;) So if your suggestion still stands, I'll add `name` parameter.

glorpen:
@domcleal Param `letsencrypt::certonly::letsencrypt_command` defaults to `$letsencrypt::command`, in this PR `$letsencrypt::command` is constructed from `$package_command` when in package mode, so it should work as expected... or am I not seeing something?

@danzilio While this PR was created, GitHub had some problems with cache/load balancing/something, so my guess is that after next change tests will run normally.

manifests/install.pp
domcleal: > In the adding relationship case, wouldn't it be expected to have working Package[$my_pkg] when providing $package_name, eg. class {'letsencrypt': package_name => $my_pkg}? After all it is named "package_name".

Yeah, if you're passing in package_name then I guess you know it - but if it's derived from params.pp then you might not.

It's probably personal preference.

domcleal:
> Param letsencrypt::certonly::letsencrypt_command defaults to $letsencrypt::command, in this PR $letsencrypt::command is constructed from $package_command when in package mode, so it should work as expected... or am I not seeing something?

Ah yes, thanks for the explanation! I missed that $command was set to $package_command.  Looks good!

PR #35, opened 2016-05-25
Change EL7 package and command to certbot

domcleal:
certbot has replaced letsencrypt in EPEL7:
https://apps.fedoraproject.org/packages/certbot
https://apps.fedoraproject.org/packages/letsencrypt


danzilio:
Thanks @domcleal!

PR #36, opened 2016-05-29
Various smaller fixes.

thomasvs:


danzilio:
Thank you for the housekeeping @thomasvs!! It's hard to keep on top of this stuff!!

PR #38, opened 2016-06-15
Fix validation of letsencrypt_command

domcleal:


danzilio:
thanks for catching this!

Issue #28, opened 2016-04-18
Using certonly in webroot mode fails with "can't convert Enumerable::Enumerator into Array"


akomakom:
Puppet: **3.8.5**
Letsencrypt: **0.5.0** via GIT (0.1.0, which is default, doesn't like Centos 6 with Ruby 1.8.7, and adding epel doesn't seem to get me any packages, thus I'm using git.)
Ruby: **2.1.2p95** on the master, **1.8.7** on the node.

```
 letsencrypt::certonly{"My letsencrypt":
    domains => ['www.my.org', 'my.org'],
    plugin => 'webroot',
    webroot_paths => ['/var/www/html/something/public']  # I also tried specifying it twice but saw use of cycle() 
  }
```

Error:

> Error: Could not retrieve catalog from remote server: Error 400 on SERVER: Failed to parse inline template: can't convert Enumerable::Enumerator into Array at /etc/puppet/modules/letsencrypt/manifests/certonly.pp:47 

Now, I'm not sure what exactly is Enumerable here, so I tried it on command line and **it works fine**:

> irb(main):001:0> webroot_paths = ['/some/path']
> => ["/some/path"]
> irb(main):002:0> domains = ['www.my.org','my.org']
> => ["www.my.org", "my.org"]
> irb(main):003:0> domains.zip(webroot_paths.cycle).map { |domain| "--webroot-path #{domain[1]} -d #{domain[0]}"}.join(" ")                                      
> => "--webroot-path /some/path -d www.my.org --webroot-path /some/path -d my.org"
>                           

Obviously I had to drop the '@' signs.

I'm not that familiar with rake and couldn't get the tests to work, so this remains unsolved on my end.

akomakom:
removing ".cycle" from letsencrypt init.pp and using a matching-size array of web roots makes this work... not sure why cycle() doesn't work the same way in puppet as it does in irb.

tomgillett:
I can +1 this issue and fix.

**Master** Puppet 3.7.5, Ruby 1.8.7
**Agent** Puppet 3.4.3, Ruby 2.1.5p273

danzilio:
@akomakom just want to confirm: you're running 1.8.7 on the agent but 2.1.2 on the master? It would make sense with 1.8.7 on the master. This module isn't tested against 1.8.7 and, to be quite honest, I have no desire to support such an old Ruby version.

akomakom:
[...]

PR #40, opened 2016-06-15
Replace webroot_paths.cycle for Ruby 1.8.7 compatibility

domcleal:
Enumerable#cycle is only available in 1.9+, so change the zipping of
domains to webroot_paths with a simplified expansion of webroot_paths
to match the number of domains.

Fixes GH-28


domcleal:
Closing in favour of #42.

PR #42, opened 2016-06-16
Remove webroot_paths cycling to match domains list

danzilio:
When fewer webroot_paths than domains are supplied, certbot/LE will use
the last webroot path given rather than the module cycling through
webroot_paths. When webroot_paths is one element, there will be no
behaviour difference.

Fixes GH-28


domcleal:
:+1: thanks!

PR #49, opened 2016-08-18
Certbot

cpitkin:
fixed naming and venv parameter that is not needed when installing from vcs. I have not tried installing from packages. 

i can also fix the spec files if you would like but it will take me a day or so to get that update added to this PR.


TJM:
The only issue that keeps catching me is.. on a new install, it tries to create the `/etc/letsencrypt/cli.ini` before `/etc/letsencrypt` exists (before the package is installed). I would suggest either managing the directory /etc/letsencrypt or adding a dependency on the "install"

cpitkin:
i can take a look and add a require if needed. Are you installing via package or vcs? I will make sure the fix works on both but want to make sure I cover your use case.

TJM:
I am using the package on CentOS 7. I honestly think it may be a problem with line 44 of config.pp ... create_ini_settings().. I would think that the "automatic" dependency would be there to ensure the directory exists, but apparently the way that its creating the file resource is somehow bypassing common sense...

Maybe we can add a "require" to the settings? or maybe a "before" on the package (and VCS install) ?

cpitkin:
looking at the code it just needs a require added to the class. the `$config_file` gets pulled from the `::certbot::params` class anyway. since this is the first time the param is used it needs to be required as a dependency since you can't write ini configs to a file that is not there. i can get that added to the PR later today or tomorrow. if you have ready access to a CentOS box can you test it out to make sure it works. all my systems are Ubuntu 14.04.

TJM:
https://tickets.puppetlabs.com/browse/MODULES-3802 <-- yes, and its a workaround for that issue

TJM:
I will test it soon, but I do have concerns... I would have thought you would require the directory (i.e. `dirname($config_file)`) ... NOTE its hardcoded above too, so it should probably be fixed there too.

However, if you need the "file" to be there, you will probably have to do the dependency in a more complicated way... because you are going to want to be sure the "installation" has finished (VCS or PKG)  "before" File[$config_file] ... and probably define the resource so that the dependency works? :-/

cpitkin:
i see what you mean and i should have spotted that sooner. i should be able to get to that this weekend sometime. I have an idea of how to fix it.

TJM:
As suspected:

```
Could not retrieve catalog from remote server: Error 400 on SERVER: Invalid relationship: Ini_setting[/etc/letsencrypt/cli.ini server] { require => File[/etc/letsencrypt/cli.ini] }, because File[/etc/letsencrypt/cli.ini] doesn't seem to be in the catalog
```

I will revert my local repo (if I can figure out how) for now :)

cpitkin:
@TJM the latest changes should fix the dependency issues. all tests now pass.

PR #50, opened 2016-08-30
Update certonly.pp

ashish1099:
Need include letsencrypt to work with puppet 4

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -38,6 +38,9 @@ define letsencrypt::certonly (
   $manage_cron          = false,
   $cron_success_command = undef,
 ) {
+
+  include ::letsencrypt
+  
   validate_array($domains)
   validate_re($plugin, ['^apache$', '^standalone$', '^webroot$'])
   if $webroot_paths {


danzilio:
Hi @ashish1099! Thanks for the PR.

This behavior is deliberate and desired. Since the `letsencrypt` class is parameterized and many users will pass those parameters we don't `include letsencrypt` here to avoid duplicate resource definitions.

This is not particularly obvious from the documentation so I'll make sure to update the docs with this information.

ashish1099:
@danzilio yes right, but it does not work with puppet 4 as what we are using here.
With puppet 4 the global variable is not reachable, so we need to include the init class.
I'm happy with any other solution as well

```
    # Install letsencrypt client packages
    class { '::profile::letsencrypt' :
      email   => 'ca-letsencrypt@test.com',
      epel    => false,
      require => File['/opt/letsencrypt'],
    }

    # https://wiki.enableit.dk/bin/view/Main/ManuallyRenewingLetSEncryptCertificates
    letsencrypt::certonly { 'test' :
      domains             => [ 'test.com', 'test.org' ],
      plugin              => 'webroot',
      webroot_paths       => ['/var/www/html'],
      manage_cron         => true,
      letsencrypt_command => 'certbot',
      additional_args     => [ '--rsa-key-size 4096', '--config-dir /opt/letsencrypt', '--text' ],
    }
  }
```

ashish1099:
and here is the class profile::letsencyrpt
```
class profile::letsencrypt (
  $email,
  $epel           = false,
  $version        = 'v0.8.0',
  $install_method = 'package',
) {
  class { '::letsencrypt' :
    configure_epel => $epel,
    email          => $email,
    install_method => $install_method,
    version        => $version,
  }
}
```

danzilio:
What version of Puppet 4 are you running?

ashish1099:
[root@puppet01 ~]# puppetserver --version
puppetserver version: 2.5.0

which gets puppet 4.6

danzilio:
Is it possible you're hitting this bug? https://tickets.puppetlabs.com/browse/PUP-6608

ashish1099:
Thanks @danzilio I will test it tomorrow and will give you an update.

ashish1099:
@danzilio we have got puppet 4.6.1 itself.
and I don't see duplicate issue even though I have called the class and the definition together

danzilio:
That's because of how the `include` function works.

If `class { 'letsencrypt': }` is evaluated before `include letsencrypt` there will be no conflict. 

If `include letsencrypt` is evaluated before `class { 'letsencrypt': }` there will be a conflict.

Since we can't guarantee the evaluation order we let the user deal with the `letsencrypt` class.

PR #51, opened 2016-08-30
run initial exec just once

Wayneoween:
I think, this exec should not run each time puppet runs.

I might be missing an use case here, so if that is the case, please enlighten me.
In my case, this exec would run with each puppet run, calling `letsencrypt --agree-tos certonly -a standalone -d $DOMAIN` and then this ncurses window would try to ask what to do and break the run.

I worked around by using `additional_args => ['--keep-until-expiring']`. It gets added a second time to the cronjob, but that doesn't do any harm.

This might be fixed in the certbot branch, but that is not officially released yet.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -68,6 +68,7 @@ define letsencrypt::certonly (
     environment => concat([ $venv_path_var ], $environment),
     creates     => $live_path,
     require     => Class['letsencrypt'],
+    refreshonly => true,
   }
 
   if $manage_cron {


danzilio:
Hi @Wayneoween. Sorry for the long lead time on this. This exec shouldn't be running every time, the exec should be idempotent based on the `creates` parameter.

pulecp:
I don't agree with this change. You want to run the initialization until it creates some file. The initialization can fail for any reason and you want from puppet to try to initialize again.

There should be option like:

    creates     => <some_file_which_is_created_after_initialization>

NOTE: @danzilio @bastelfreak This issue can be closed because 'refreshonly  => true' option is there since initial commit of _init.pp_ :). See (https://github.com/voxpupuli/puppet-letsencrypt/commit/12019701ec328a3e8ed152ef4eb8a6f0f7524467#diff-60ae41fd0a31977447947f59940ee9a4R41)

Another issue should be created to switch from 'refreshonly => true' to 'creates => <some_generated_file>'

PR #60, opened 2016-10-17
Fix cronspam

jethrocarr:
Added `--quiet` argument to certbot. This implies `--non-interactive` so immediately stops getting exceptions trying to load dialog boxes, but also ensures that we don't send an email every day saying no certs for renewal.

`--quiet` will still result in output if certbot outputs anything on STDERR.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -71,7 +71,7 @@ define letsencrypt::certonly (
   }
 
   if $manage_cron {
-    $renewcommand = "${command_start}--keep-until-expiring ${command_domains}${command_end}"
+    $renewcommand = "${command_start}--keep-until-expiring --quiet ${command_domains}${command_end}"
     if $cron_success_command {
       $cron_cmd = "${renewcommand} && (${cron_success_command})"
     } else {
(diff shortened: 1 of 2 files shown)


hdeadman:
--quiet option doesn't appear to be present if you are installing the package on Ubuntu 16.04, which is the default for the module. The resulting error breaks the cronjob. 

The latest package for Ubuntu 16.04 appears to be: 

letsencrypt                      0.4.1-1

jethrocarr:
Bugger @hdeadman  - I tested on CentOS 7 which ships 0.8.1 (via EPEL repo). Could wrap it inside a per-distro conditional to disable the flag on older distributions... although there's probably a question around what version of certbot should be considered the minimum?

hdeadman:
@jethrocarr - Well I will probably have to switch to vcs instead of package. If you want to declare a minimum version then the default for Ubuntu 16.04 (Long Term Support) should probably be vcs and for 16.10 it should be package with the package_name of certbot. Looks like they have certbot 0.8.1-2 on 16.10. 

[16.10](http://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=certbot&searchon=names)

siebrand:
The "--quiet" option was added in certbot [0.14.0 released on 2017-05-04](https://github.com/certbot/certbot/blob/master/CHANGELOG.md)

PR #70, opened 2016-12-21
Bump puppet minimum version_requirement to 3.8.7

juniorsysadmin:
We currently only run automated tests against Puppet 3 latest and
therefore cannot guarantee that this module works as is expected with
earlier Puppet 3 versions


juniorsysadmin:
Done in a modulesync

PR #74, opened 2016-12-30
Removing --quiet from certonly cron

craigwatson:
Newer version of LetsEncrypt do not use the `--quiet` parameter:

```letsencrypt: error: unrecognized arguments: --quiet```

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -71,7 +71,7 @@ define letsencrypt::certonly (
   }
 
   if $manage_cron {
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
@craigwatson Corresponding tests in spec/ need to be changed as well

craigwatson:
@juniorsysadmin Good catch, thanks - tests updated and tests squashed :)

pschichtel:
@craigwatson which version exactly doesn't support the --quiet flag anymore? Version 0.14.0 on my Debian machine supports it

PR #76, opened 2017-01-02
fix(packages): Correct package name for Debian 8

lvicainne:
Hello,
I noticed on my Debian machine (Debian 8), I should use the package certbot instead of Letencrypt.
What do you think ?
Thanks,
Regards,

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -14,10 +14,10 @@ class letsencrypt::params {
     'server' => 'https://acme-v01.api.letsencrypt.org/directory',
   }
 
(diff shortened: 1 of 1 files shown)


bastelfreak:
Hi @lvicainne, thanks for the PR. Can you take a look at the failed travis tests?

ekohl:
According to https://packages.debian.org/jessie-backports/certbot this should be correct. I can prepare a PR that includes fixed tests.

lvicainne:
Thanks for #88 ! +1 
I close this one :-)

PR #77, opened 2017-01-02
Create_resource for certonly

lvicainne:
Hello,
I propose to add create_resource in order to be better compliant with Hiera,
Thanks.
Regards

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -46,6 +46,8 @@
 #   A flag to agree to the Let's Encrypt Terms of Service.
 # [*unsafe_registration*]
 #   A flag to allow using the 'register-unsafely-without-email' flag.
+# [*certonly*]
+#   A hash containing all the configuration for creating a certonly
 #
 class letsencrypt (
   $email               = undef,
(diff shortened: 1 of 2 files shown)


bastelfreak:
Hi @lvicainne thanks for the PR!

---

@igalic do you have any objections here on using create_resources? AFAIK you're not a big fan of it. Any ideas for a better implementation?

manifests/init.pp
```diff
@@ -103,4 +106,8 @@
     environment => concat([ "VENV_PATH=${venv_path}" ], $environment),
     refreshonly => true,
   }
+
+  $real_certonly = hiera_hash('letsencrypt::certonly', $certonly)
```
baurmatt: Please don't do Hiera lookup with the module

manifests/init.pp
```diff
@@ -103,4 +106,8 @@
...
+  create_resources(::letsencrypt::certonly, $real_certonly)
```
baurmatt: Please use `each` as it's Puppet 4 native.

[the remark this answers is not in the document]
bastelfreak: All of our modules are only tested on Puppet 4.10 an newer: https://github.com/voxpupuli/puppet-letsencrypt/blob/master/metadata.json#L58

ody:
Good day @lvicainne. We certainly appreciate your contribution!

Doing house cleaning and due to inactivity I am going to close this PR. If you are still interested in the proposed functionality then please re-open with previous comments addressed.

Thanks again!

PR #84, opened 2017-02-12
Added custom_plugin param in certonly to disable use of '-a' flag.

Lavaburn:
I need to validate my domain using DNS, which I found is possible with external_auth plugin. 
Using this plugin with certbot however requires '-a <plugin>' flag to be removed from the certbot command.

This fix introduces a new parameter and does not break any existing code. As an extra, I allowed nginx as plugin, which is also supported now in certbot.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -7,6 +7,8 @@
 #
 # [*domains*]
 #   Namevar. An array of domains to include in the CSR.
+# [*custom_plugin*]
+#   Whether to use a custom plugin in additional_args and disable -a flag.
 # [*plugin*]
 #   The authenticator plugin to use when requesting the certificate.
 # [*webroot_paths*]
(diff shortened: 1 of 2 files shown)


manifests/certonly.pp
```diff
@@ -32,6 +34,7 @@
 #
 define letsencrypt::certonly (
   $domains              = [$title],
+  $custom_plugin        = false,
```
bastelfreak: can you use the new Pupept 4 Datatype Boolean instead of the validate_re down below?

oc243:
Thanks @Lavaburn for writing this PR. It fixes the problem I was just looking into :). I hope it can be merged soon ;-)

bastelfreak:
Hi @Lavaburn, could you please rebase against latest master to fix the merge conflict?

bastelfreak:
thanks!

PR #87, opened 2017-04-26
fix "-":12: command too long. errors in crontab file, can't install.

saimonn:
When handling a certificate with >50 altnames with `$manage_cron = true`, the cron command generated was too long to be installed by puppet:
```
Notice: /Stage[main]/Profiles[...]/Letsencrypt::Certonly[default]/Cron[letsencrypt renew cron default]/ensure: created
"-":12: command too long
errors in crontab file, can't install.
Notice: Applied catalog in 28.44 seconds
```

This PR proposes to move the renew command in a script, and call this script with the cron daemon.

NB:  puppet run didn't fail on this crontab error, but this is another problem.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -94,8 +94,15 @@ define letsencrypt::certonly (
     }
     $cron_hour = fqdn_rand(24, $title) # 0 - 23, seed is title plus fqdn
     $cron_minute = fqdn_rand(60, $title ) # 0 - 59, seed is title plus fqdn
+    file { "/usr/local/sbin/letsencrypt-renew-cron-${title}":
+      ensure  => 'file',
+      mode    => '0755',
+      owner   => 'root',
+      group   => 'root',
+      content => "#!/bin/sh\n${cron_cmd}",
+    }
     cron { "letsencrypt renew cron ${title}":
-      command     => $cron_cmd,
+      command     => "/usr/local/sbin/letsencrypt-renew-cron-${title}",
       environment => concat([ $venv_path_var ], $environment),
       user        => root,
       hour        => $cron_hour,
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -94,7 +94,7 @@ define letsencrypt::certonly (
     }
     $cron_hour = fqdn_rand(24, $title) # 0 - 23, seed is title plus fqdn
     $cron_minute = fqdn_rand(60, $title ) # 0 - 59, seed is title plus fqdn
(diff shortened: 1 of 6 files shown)


mcanevet:
LGTM

oranenj:
Honestly, this sounds like a bug in cron, but yeah... I guess a workaround is needed. 

I don't like the fact that this creates an unbounded number of files in /usr/local/sbin that might not be cleaned up if something changes, though.

In puppet-selinux we used the `puppet_vardir` fact from stdlib to create a directory in the puppet agent var directory and files that we could then safely purge. Maybe that is a neater solution?

saimonn:
Neater idea yes, I've pushed a commit to use `${puppet_vardir}/letsencrypt/` directory instead.

oranenj:
@saimonn thanks! However, it should be validated using the Stdlib::Absolutepath type instead of just a string, and since the directory is purged, it would be a good idea to force it to have a hardcoded segment. Something like

Stdlib::Absolutepath $script_root = $::puppet_vardir,

Then 
```
file {"${script_root}/puppet-letsencrypt":
  ensure  => directory,
  purge   => true,
  recurse => true,
}
```

And then use that; if someone says their $script_root is /etc, they won't be too sad :)
Might be I'm overly cautious, but I've found that it's usually better in this line of work.

saimonn:
The hardcoded segment part is already [defined in params class](https://github.com/voxpupuli/puppet-letsencrypt/pull/87/files#diff-2f383a197b1a428561b364755701d60eR12) (`"${::puppet_vardir}/letsencrypt"`).

I agree we should use typed parameters instead of old `validate_*` functions, but isn't this migration out of scope of this PR ?

manifests/certonly.pp
```diff
@@ -94,8 +94,15 @@
...
+    file { "${letsencrypt::cron_scripts_path}/renew-${title}.sh":
+      ensure  => 'file',
+      mode    => '0755',
+      owner   => 'root',
+      group   => 'root',
+      content => "#!/bin/sh\n${cron_cmd}",
+    }
     cron { "letsencrypt renew cron ${title}":
-      command     => $cron_cmd,
+      command     => "${::puppet_vardir}/letsencrypt/renew-${title}.sh",
```
yastupin: shouldn't `command` be `${letsencrypt::cron_scripts_path}/renew-${title}.sh`?

saimonn: That's right, I replaced this with `${::letsencrypt::cron_scripts_path}/renew-${title}.sh` instead.

PR #88, opened 2017-04-29
Use the correct package names on Debian

ekohl:
This also allows using jessie-backports. Replaces https://github.com/voxpupuli/puppet-letsencrypt/pull/76.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,21 @@ The module can integrate with [stahnma/epel](https://forge.puppetlabs.com/stahnm
 to set up the repo by setting the `configure_epel` parameter to `true` (the default for RedHat) and
 installing the module.
 
+On Debian Jessie the module assumes the package certbot is available. This
+package can be found in jessie-backports. When using
+[puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) the following code
+can be used:
+
+```puppet
+include ::apt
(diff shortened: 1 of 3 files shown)


ekohl:
I also used the following apt config to pull the right packages in. Not sure it belongs in this module, maybe just in the README?

```puppet
include ::apt
include ::apt::backports
apt::pin { 'jessie-backports-letsencrypt':
  release  => 'jessie-backports;,
  packages => prefix(['acme', 'cryptography', 'openssl', 'psutil', 'setuptools', 'pyasn1', 'pkg-resources'], 'prefix-'),
  priority => 700,
}
```

oranenj:
@ekohl: Can you add that to the README? (with syntax fixed, please :))

What's the 'prefix-' for though?

ekohl:
The ```prefix()``` comes from stdlib. The actual package names are python-acme, python-cryptography but this resulted in a shorted line so puppet-lint is happy. I can go with the explicit version if that would be preferred.

oranenj:
@ekohl: In your code snippet you were adding a 'prefix-' prefix to the packages instead of 'python-' though :)

Using `prefix()` in the example is fine.

ekohl:
@oranenj you're totally right. I guess late night hacking results in bad code sometimes :)

PR #90, opened 2017-05-22
Add parameters to customize cron's hour and minute

maxenced:
When having many certificates managed by LE on the same host, the random function to set minutes and hours does not work that well. As a consequence, errors are raised when some "race condition" happens (in log file rotation for ex).

Also, one may want to restrict the time where cron can be run (only during night for ex).

This PR add 2 parameters to specifically set cron's hour and minute. Having them enables easy customization of every cron run time, without having to fully disable the manage_cron stuff and manage everything (command line etc ...) yourself.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -29,6 +29,10 @@
 # [*cron_success_command*]
 #   String representation of a command that should be run if the renewal command
 #   succeeds.
+# [*cron_hour*]
+#   If set, force cron's hour to this value. Must be between 0 and 23.
+# [*cron_minute*]
+#   If set, force cron's minute to this value. Must be between 0 and 59.
 #
 define letsencrypt::certonly (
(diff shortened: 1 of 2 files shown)


maxenced:
Hi, looks like master is now switched to support puppet4 only . Should I put this MR against puppet3 branch ?

bastelfreak:
Hi  @maxenced, thanks for the PR and sorry for the late reply. Puppet 3 is end of life since several months, so we don't want to support it anymore. Can you rebase against our latest master, to fix the merge conflicts? Afterwards we can merge it and make a new release.

Dan33l:
the current version of module looks to have this improvement:
https://github.com/voxpupuli/puppet-letsencrypt/blob/cb60d952a624fffcba221784e758ba61eb26874c/manifests/certonly.pp

PR #93, opened 2017-06-11
replace validate_* with proper puppet4 datatypes

bastelfreak:


bastelfreak:
We should merge https://github.com/voxpupuli/puppet-letsencrypt/pull/92 first

PR #96, opened 2017-07-28
fixed randomness if the domain is almost same

ashish1099:
like example.com and www.example.com, having same cron timing which was getting failed, because certbot can't have two instance at the same time


ashish1099:
any update on this one ?

bastelfreak:
Thanks for this PR @ashish1099!

PR #109, opened 2017-11-26
bump inifile dependency to allow 2.x

bastelfreak:

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -52,7 +52,7 @@
     },
     {
       "name": "puppetlabs/inifile",
-      "version_requirement": ">= 1.4.1 < 2.0.0"
+      "version_requirement": ">= 1.4.1 < 3.0.0"
     },
     {
       "name": "puppetlabs/vcsrepo",
--- a/metadata.json
+++ b/metadata.json
@@ -52,7 +52,7 @@
     },
     {
       "name": "puppetlabs/inifile",
-      "version_requirement": ">= 1.4.1 < 3.0.0"
+      "version_requirement": ">= 2.0.0 < 3.0.0"
     },
     {
       "name": "puppetlabs/vcsrepo",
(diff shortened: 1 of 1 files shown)


metadata.json
```diff
@@ -52,11 +52,11 @@
     },
     {
       "name": "puppetlabs/inifile",
-      "version_requirement": ">= 1.4.1 < 2.0.0"
+      "version_requirement": ">= 2.0.0 < 3.0.0"
```
ekohl: Why increase the lower bound as well? The major version bump was because of dropping puppet 3 but that's OK for us.

metadata.json
```diff
@@ -52,11 +52,11 @@
...
     },
     {
       "name": "puppetlabs/vcsrepo",
-      "version_requirement": ">= 1.3.1 < 3.0.0"
+      "version_requirement": ">= 2.0.0 < 3.0.0"
```
ekohl: Same here

metadata.json
bastelfreak: older ones don't support puppet5

metadata.json
bastelfreak: older ones don't support puppet5

metadata.json
ekohl: I'd say this is still compatible and it's up to the dependency resolver to figure out but I'm not sure what's the best practice here.

pschichtel:
What is the state here?

metadata.json
juniorsysadmin: I'm pretty sure that the dependency resolver is kinda dumb and doesn't take into account Puppet versions. Also, people consume/fetch modules in different ways not necessarily through `puppet module install`. So in my opinion having a common lower bound for all dependencies is the right thing to do.

Issue #66, opened 2016-12-07
apache virtualhost and letsencrypt cert


hoffereka:
I am stuck with a problem, please help.
I want to install an apache with a letsencrypt generated cert, using your puppet module.
My problem is, apache wont start wit missing cert, but letsencrypt wont get cert without apache.
How can I do this with puppet in one pass:
- start apache without ssl vhosts
- get and install letsencrypt cert
- add ssl virtualhosts and restart apache

Tnaks in advance.

thiagomarinho:
@hoffereka I have the same issue, did you solve it?

ScottWales:
I've worked around this with a custom fact that returns the live certificate directories:
```ruby
require "pathname"

Facter.add(:letsencrypt_live) do
    setcode do
        Pathname.new("/etc/letsencrypt/live").children.select(&:directory?).collect(&:basename)
    end
end
```

then in puppet only enabling the ssl server when it's in the list of certificates:
```puppet
::apache::vhost {"${hostname} nossl": 
    # ...
}
::letsencrypt::certonly {$hostname:
    require => Service[httpd],
    # ...
}
if $hostname in $::letsencrypt_live {
    ::apache::vhost{"${hostname} ssl":
        # ...
    }
}
```
This does require two runs of Puppet before the SSL host is live

benjunmun:
I've got it working in a single run, but it's not pretty. This sample code is for ubuntu 16.04. The key parts are running letsencrypt between the ssl and non-ssl vhosts, and restarting apache - the puppet default mpm is different from the ubuntu package's default mpm.

    class { ::letsencrypt:
        config     => {
            email  => 'webmaster@domain.example',
            server => 'https://acme-staging.api.letsencrypt.org/directory',
        }
    }

    class { ::apache:
        default_vhost => false,
    }

    #This fixes:
    #   AH00534: apache2: Configuration error: The MPM cannot be changed during restart.
    #The letsencrypt plugin only sends a SIGUSR1, which is insufficient
    exec {'Restart Apache for MPM changes':
        command     => '/bin/systemctl restart apache2.service',
        refreshonly => true,
    }

    Apache::Vhost<|ssl == false|> ~> Exec['Restart Apache for MPM changes'] -> Letsencrypt::Certonly<||> -> Apache::Vhost<|ssl == true|>

    package {'python-letsencrypt-apache':
        ensure => present,
        before => Letsencrypt::Certonly['test.domain.example'],
    }
    
[...]

PR #110, opened 2017-12-09
letsencrypt_directory, letsencrypt_lookup

cwells:
This PR addresses issue #66.  See comments for discussion.


bastelfreak:
@cwells thanks for the PR. Can you check the used email address in your commits? It isn't associated with your github account.

lib/facter/letsencrypt_directory.rb
```diff
@@ -0,0 +1,29 @@
...
+        cert = OpenSSL::X509::Certificate.new(File.new(pem).read)
+        san = cert.extensions.find { |e| e.oid == 'subjectAltName' }
+        names = san.value.split(',').map { |entry| entry.split(':')[1] }
+        names.each do |n|
+          certs[n] = path.to_s
+        end
+      end
+    end
+
+    certs
```
bastelfreak: should we instead use a structured fact? Would make the function obsolete or easier

cwells: Amending git history is a PITA. Unless there's some reason it matters, I'd rather just leave the email as-is. 

The fact is structured (it's a hash). While you can certainly index into it directly, the function also handles wildcards, which isn't something a structure is going to provide.

thejambavan:
Please could this PR be merged? These facts would come in quite handy

igalic:
I like this

the code looks fine (on a first glance, i haven't tested it yet) too
what's wrong with merging this and seeing what happens (before releasing)?

cwells:
Closing due to inactivity.

PR #114, opened 2018-02-27
Add OpenBSD support

arthurbarton:
Low touch OpenBSD support.

Required adding a new parameter for the file group for cronjob management.

There is a failing test, but it fails against current master also.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -47,9 +47,8 @@ class letsencrypt::params {
     $configure_epel = false
   }
 
-  if $::osfamily == 'OpenBSD' {
-    $cron_owner_group = 'wheel'
-  } else {
-    $cron_owner_group = 'root'
+  $cron_owner_group = $::osfamily ? {
+    'OpenBSD' =>  'wheel',
+    default   =>  'root',
   }
 }
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -31,6 +31,10 @@ class letsencrypt::params {
     $install_method = 'package'
     $package_name = 'app-crypt/certbot'
     $package_command = 'certbot'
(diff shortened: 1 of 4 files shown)


manifests/params.pp
```diff
@@ -42,4 +46,10 @@
   } else {
     $configure_epel = false
   }
+
+  if $::osfamily == 'OpenBSD' {
```
bastelfreak: can you use a selector here?
https://puppet.com/docs/puppet/5.3/lang_conditional.html#selectors

manifests/params.pp
```diff
@@ -31,6 +31,10 @@
     $install_method = 'package'
     $package_name = 'app-crypt/certbot'
     $package_command = 'certbot'
+  } elsif $::osfamily == 'OpenBSD' {
```
bastelfreak: this whole block should probably be changed to a switch/case statement. You don't need to convert it, but feel free if you want to.

ekohl: Might also be a good candidate for hiera data files but that's probably out of scope for this PR.

bastelfreak:
@arthurbarton can you take a look at the failing travis runs? Afterwards I'm happy to merge it.

arthurbarton:
travis fixed, swapped the osfamily back to the classic fact. Not sure why that makes a difference

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-letsencrypt",
-  "version": "2.2.1-rc0",
+  "version": "2.2.1-rc1",
```
bastelfreak: that isn't needed but also doesn't hurt anybody. Our release process will bump this to a non-rc version.

PR #132, opened 2018-08-20
Remove using package for 16.04, as it installs package 0.4.1-1 instea…

### LINKED ISSUE #131: Ubuntu 16.04 installs python-letsencrypt 0.4.1-1 by default instead of 0.9.3
## Affected Puppet, Ruby, OS and module versions/distributions

- Distribution: Ubuntu
- Module version: 2.4.0

## How to reproduce (e.g Puppet code you use)
Install on Ubuntu-16.04

## What are you seeing
The default letsencrypt package installed is python-letsencrypt 0.4.1-1 https://packages.ubuntu.com/xenial/python-letsencrypt

## What behaviour did you expect instead
It should use version 0.9.3 by default

## Workaround
install_method       => 'vcs',

GrayPockets:
Removes Ubuntu 16.04 exception (which causes it to use the default of $install_method = 'vcs')

Fixes #131

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -19,11 +19,6 @@ class letsencrypt::params {
     $package_name = 'certbot'
     $package_command = 'certbot'
     $config_dir = '/etc/letsencrypt'
-  } elsif $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '16.04') >= 0 {
-    $install_method = 'package'
-    $package_name = 'letsencrypt'
-    $package_command = 'letsencrypt'
-    $config_dir = '/etc/letsencrypt'
   } elsif $::osfamily == 'RedHat' and versioncmp($::operatingsystemmajrelease, '7') >= 0 {
     $install_method = 'package'
     $package_name = 'certbot'


bastelfreak:
Hi @GrayPockets, thanks for the patch. Can you please add unit tests so we know it is working as expected on Ubuntu and rebase against our master?

Dan33l:
For all OSes, the module install the version of `letsencrypt` provided by the distrib. If it is not the desired version, it is possible to install `letsencrypt` from git repository. I don't understand why Ubuntu would be different from others OSes.

IMO it is not a bug. And i understand the workaround given in the issue as a normal behavior.

Dan33l:
It exists a bug between supported version in metadata.json and params.pp  . I agree that these two files be updated synchronously.

manifests/params.pp
```diff
@@ -19,11 +19,6 @@
     $package_name = 'certbot'
     $package_command = 'certbot'
     $config_dir = '/etc/letsencrypt'
-  } elsif $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '16.04') >= 0 {
```
Dan33l: Instead of removing the condition, we may consider package from ubuntu 18.04 and upper enough up to date by default.
```
} elsif $::operatingsystem == 'Ubuntu' and versioncmp($::operatingsystemrelease, '18.04') >= 0 {
```
We probably should have to add Ubuntu 16.04 and 18.04 as supported OSes in `metadata.json`.

ekohl:
@GrayPockets could you have a look at the comments?

Dan33l:
closed since #158 merged.

ekohl:
@Dan33l Adding 18.04 is different from using 16.04 installing via VCS.

Dan33l:
@Dan33l i read too fast. You are right.

ekohl:
https://bugs.launchpad.net/ubuntu/xenial/+source/python-letsencrypt/+bug/1640978/comments/73 states that a certbot update in Xenial is planned. This will become more important when the TLS-SNI-01 validation option is turned off by LE (which will happen soon).

Dan33l:
The version provided by ubuntu 16.04 is now python-letsencrypt - 0.7.0
https://bugs.launchpad.net/ubuntu/xenial/+source/python-letsencrypt/+bug/1640978/comments/129

Dan33l:
Since package used by Ubuntu 16.04 is updated, i proceed to close the ticket.

Feel free to reopen if you find it should be usefull.

PR #137, opened 2018-09-04
Bug/proper domain validation

### LINKED ISSUE #94: Adding extra domain fails
Using 'letsencrypt::certonly' it appears that adding extra domain names does not work once the certificate has been issued.

Running the command that eventually happens reveals that the certbot code goes user-interactive and asks if you wish to expand the certificate to cover the new domains, but the puppet module does not deal with this and the current certificate remains in place.

I have had a look on the letsencrypt community board and issue tracker, it seems that this is still something undergoing change but there is an '--expand' flag that certbot should honour to add new domains, possibly ONLY if they are a complete superset (you can't remove any).

I'm not sure what a good solution is here, possibly adding '--expand' for now but it will probably still fail when domains are removed since doing this does not affect the currently issued certificate and does not change the renewal paramters in /etc/letsencrypt/renewal/<DOMAIN>.conf

baurmatt:
This PR contains two commits:

1. Refactoring:
```
Remove inline_template in favour of native Puppet code

* Migrates all inline_template calls to Puppet native code, as this is
recommended best practice
* Changes the default value/data type of webroot_paths and
additional_args to an Array instead of Optional[Array]
* Adds a data type for the certificate plugins for better readability
```

2. Bug fix for expanding a certificate
```
Allow additional domains to be added later

This commit adds support for expansion/shrinking of an already existing
certificate by defining a unique identifier (--cert-name) which maps to
the Puppet resource title. By doing this, we can ensure that always all
requested domains are contained in the certificate.
```

Fixes #94

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -33,24 +33,22 @@
 #   succeeds.
 #
 define letsencrypt::certonly (
-  Array $domains                                            = [$title],
-  Boolean $custom_plugin                                    = false,
-  Enum['apache', 'standalone', 'webroot', 'nginx', 'dns-route53', 'dns-google'] $plugin  = 'standalone',
-  Optional[Array] $webroot_paths                            = undef,
-  String $letsencrypt_command                               = $letsencrypt::command,
-  Optional[Array] $additional_args                          = undef,
-  Array $environment                                        = [],
-  Boolean $manage_cron                                      = false,
-  Boolean $suppress_cron_output                             = false,
-  $cron_before_command                                      = undef,
-  $cron_success_command                                     = undef,
-  Stdlib::Unixpath $config_dir                              = $letsencrypt::config_dir,
+  Array[Stdlib::Host]     $domains              = [$title],
+  Boolean                 $custom_plugin        = false,
+  Letsencrypt::Plugin     $plugin               = 'standalone',
+  Array[Stdlib::Unixpath] $webroot_paths        = [],
+  String[1]               $letsencrypt_command  = $letsencrypt::command,
+  Array[String[1]]        $additional_args      = [],
+  Array[String[1]]        $environment          = [],
+  Boolean                 $manage_cron          = false,
+  Boolean                 $suppress_cron_output = false,
+  Optional[String[1]]     $cron_before_command  = undef,
+  Optional[String[1]]     $cron_success_command = undef,
+  Stdlib::Unixpath        $config_dir           = $letsencrypt::config_dir,
 ) {
 
-  if $plugin == 'webroot' {
-    unless $webroot_paths {
-      fail("The 'webroot_paths' parameter must be specified when using the 'webroot' plugin")
-    }
+  if $plugin == 'webroot' and empty($webroot_paths) {
+    fail("The 'webroot_paths' parameter must be specified when using the 'webroot' plugin")
   }
 
   if ($custom_plugin) {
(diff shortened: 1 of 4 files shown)


types/plugin.pp
```diff
@@ -0,0 +1 @@
+type Letsencrypt::Plugin = Enum['apache', 'standalone', 'webroot', 'nginx', 'dns-route53', 'dns-google']
```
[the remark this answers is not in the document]
baurmatt: Sure, wasn't aware that this is possible oO Thanks for the hint!

manifests/certonly.pp
```diff
@@ -59,15 +57,31 @@
...
+    $_command_domains = join($domains, ' -d ')
+    $command_domains  = "-d ${_command_domains}"
   }
-  $command_end = inline_template('<% if @additional_args %> <%= @additional_args.join(" ") %><%end%>')
+
+  if empty($additional_args) {
+    $command_end = undef
+  } else {
+    # ['',] adds an additional whitespace in the front
+    $command_end = join(['',] + $additional_args, ' ')
```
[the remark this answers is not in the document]
baurmatt: Tried that, but sadly this isn't possible for strings:
```
mbaur(2018-09-04 14:34:50)  /tmp
$ cat test.pp
$test = ' ' + 'foobar'
warning($test)
mbaur(2018-09-04 14:34:51)  /tmp
$ puppet apply test.pp
Error: Evaluation Error: The value ' ' cannot be converted to Numeric. (file: /private/tmp/test.pp, line: 1, column: 9) on node mbaur.syseleven.de
```

Could find a better (readable) solution for this, that's why i decided to just add a comment which explain what's going on.

manifests/certonly.pp
```diff
@@ -33,24 +33,22 @@
...
-  Optional[Array] $additional_args                          = undef,
-  Array $environment                                        = [],
-  Boolean $manage_cron                                      = false,
-  Boolean $suppress_cron_output                             = false,
-  $cron_before_command                                      = undef,
-  $cron_success_command                                     = undef,
-  Stdlib::Unixpath $config_dir                              = $letsencrypt::config_dir,
+  Array               $domains              = [$title],
+  Boolean             $custom_plugin        = false,
+  Letsencrypt::Plugin $plugin               = 'standalone',
```
bastelfreak: yey custom datatype \o/

manifests/certonly.pp
```diff
@@ -33,24 +33,22 @@
...
+  Array               $webroot_paths        = [],
+  String              $letsencrypt_command  = $letsencrypt::command,
+  Array               $additional_args      = [],
+  Array               $environment          = [],
+  Boolean             $manage_cron          = false,
+  Boolean             $suppress_cron_output = false,
+  Optional[String]    $cron_before_command  = undef,
```
bastelfreak: can you use ```String[1]``` for all the strings?

manifests/certonly.pp
```diff
@@ -33,24 +33,22 @@
...
-  Optional[Array] $webroot_paths                            = undef,
-  String $letsencrypt_command                               = $letsencrypt::command,
-  Optional[Array] $additional_args                          = undef,
-  Array $environment                                        = [],
-  Boolean $manage_cron                                      = false,
-  Boolean $suppress_cron_output                             = false,
-  $cron_before_command                                      = undef,
-  $cron_success_command                                     = undef,
-  Stdlib::Unixpath $config_dir                              = $letsencrypt::config_dir,
+  Array               $domains              = [$title],
```
bastelfreak: can you enforce the contents of the arrays as well? Probably ```Array[String[1]]``` for most occasions.

manifests/certonly.pp
baurmatt: We love them! :)

manifests/certonly.pp
baurmatt: Sure!

manifests/certonly.pp
baurmatt: Done, but decided to use `Array[Stdlib::Host]` for `$domains` and `Array[Stdlib::Unixpath]` for `$webroot_paths`.

spec/type_aliases/plugin_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'spec_helper'
+
+describe 'Letsencrypt::Plugin' do
+  it { is_expected.to allow_values('apache', 'standalone', 'webroot', 'nginx', 'dns-route53', 'dns-google') }
```
bastelfreak: please add this:
```
it { is_expected.to compile.with_all_deps }
```

that is the minimal test that should be present in all spec files. It ensures that the catalog compiles.

types/plugin.pp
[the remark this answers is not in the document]
bastelfreak: custom datatypes \o/

spec/type_aliases/plugin_spec.rb
baurmatt: This isn't working for data types:

```
     NameError:
       undefined local variable or method `compile' for #<RSpec::ExampleGroups::LetsencryptPlugin:0x00000000055c3118>
```

ekohl: Yes, type aliases don't really compile. Testing whether it allows and disallows certain values is sufficient testing though.

PR #142, opened 2018-10-15
Allow wildcard domains for certonly::domains

Wimmesberger:

The current datatype `Array[Stdlib::Host]` for `certonly::domains` is too strict and does not allow for domains containing asterisks to generate wildcard certificates. 

My proposed change would be:
`Array[Stdlib::Host]` -> `Array[String[1]]`

Feel free to implement some other validation pattern in case `Array[String[1]]` is too loose.

Bonus: Added a case of `*.example.com` to spec test (excluding webroot plugin since you need the DNS-01 challenge)

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -33,7 +33,7 @@
 #   succeeds.
 #
 define letsencrypt::certonly (
-  Array[Stdlib::Host]     $domains              = [$title],
+  Array[String[1]]        $domains              = [$title],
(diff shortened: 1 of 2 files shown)


juniorsysadmin:
@Wimmesberger E-mail address for your commit does not match your GitHub account. It is preferable for Voxpupuli Pull Requests that it does.

Dan33l:
@Wimmesberger Thank you. 
Your PR is now release on the puppet forge.

PR #143, opened 2018-11-05
make the module compatible with FreeBSD

Wayneoween:

As FreeBSD also has no `root` group, the only change I noticed would be necessary to successfully run this module on FreeBSD is the same like for OpenBSD.


ekohl:
Thanks!

Dan33l:
@Wayneoween Thank you. 
Your PR is now release on the puppet forge.

PR #146, opened 2018-11-20
Add cron day of month parameter

brigriffin:
This PR adds support for specifying one or many day(s) of the month when to run the certbot cron. The default is to run everyday (`*` in cron syntax) and a user can now specify one or multiple days of the month when to run the certbot cron. For example some people prefer to run it twice per month instead of everyday. This is now possible with this PR.

n/a


manifests/certonly.pp
```diff
@@ -44,6 +44,7 @@
   Boolean                 $suppress_cron_output = false,
   Optional[String[1]]     $cron_before_command  = undef,
   Optional[String[1]]     $cron_success_command = undef,
+  Optional[Array]         $cron_monthday        = ['*'],
```
ekohl: Since you provide a default, this can just be `Array`. Should it also be `Array[String[1]]`?

brigriffin: Sure, we could omit the `Optional` and have just an `Array` instead.  If we use `Array[String[1]]` then one needs to specify the days in the `cron_monthday` parameter as strings like `['1', '15']`. `[1, 15]` will not work because puppet will complain with `expects a String value, got Integer`. What do you think is better? Having `Array` or `Array[String[1]]` ?

ekohl: I'm a fan of strict data types so perhaps `Array[Variant[Integer, String[1]]]`

brigriffin: I couldn't agree more with you. Shall I submit a new commit to this PR including this change?

brigriffin:
@ekohl have a quick look at 500ecf2 when you have a moment. I have now added strict data typing for the `cron_monthday` param as discussed above.

manifests/certonly.pp
```diff
@@ -33,18 +33,19 @@
...
+  Letsencrypt::Plugin                $plugin               = 'standalone',
+  Array[Stdlib::Unixpath]            $webroot_paths        = [],
+  String[1]                          $letsencrypt_command  = $letsencrypt::command,
+  Array[String[1]]                   $additional_args      = [],
+  Array[String[1]]                   $environment          = [],
+  Boolean                            $manage_cron          = false,
+  Boolean                            $suppress_cron_output = false,
+  Optional[String[1]]                $cron_before_command  = undef,
+  Optional[String[1]]                $cron_success_command = undef,
+  Array[Variant[Integer, String[1]]] $cron_monthday        = ['*'],
```
bastelfreak: Is there a known minimal/maximal value for the integer? That can also be enforced with:
```suggestion
  Array[Variant[Integer[0, 60], String[1]]] $cron_monthday = ['*'],
```

See also: https://puppet.com/docs/puppet/5.4/lang_data_number.html#the-integer-data-type

brigriffin: Well spotted! Based on the man page of crontab the integer value can be between 0-59. I will commit the change to this branch in a few minutes.

bastelfreak: thanks. I will merge it after it turned green.

brigriffin: Cheers :+1:

Dan33l:
@brigriffin Thank you. 
Your PR is now released on the puppet forge.

brigriffin:
Cheers!

PR #151, opened 2019-01-08
Fix exec onlyif logic

baurmatt:
This commit fixes the improved check logic introduces in
460a7b7cd9cca267e5cb35568ccf4cc88faa97f0. It ensure that the exec is
actually triggered if the cert file doesn't exist.

**Current master:**

* Cert doesn't exist:
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted.de/cert.pem
ls: cannot access /etc/letsencrypt/live/www.redacted.de/cert.pem: No such file or directory
root@lb:/# test -f /etc/letsencrypt/live/www.redacted.de/cert.pem && ( openssl x509 -in /etc/letsencrypt/live/www.redacted.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted.de redacted.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
1
```
==> Wouldn't run

* Cert does exists but one domain is missing
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted2.de/cert.pem
lrwxrwxrwx 1 root root 47 Dec 13 12:17 /etc/letsencrypt/live/www.redacted2.de/cert.pem -> ../../archive/www.redacted2.de/cert1.pem
root@lb:/# openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*'
DNS:redacted2.de
DNS:www.redacted2.de
root@lb:/# test -f /etc/letsencrypt/live/www.redacted2.de/cert.pem && ( openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted2.de redacted2.de foo.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
0
```
==> Would run

* Cert does exists and all domains match
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted2.de/cert.pem
lrwxrwxrwx 1 root root 47 Dec 13 12:17 /etc/letsencrypt/live/www.redacted2.de/cert.pem -> ../../archive/www.redacted2.de/cert1.pem
root@lb:/# openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*'
DNS:redacted2.de
DNS:www.redacted2.de
root@lb:/# test -f /etc/letsencrypt/live/www.redacted2.de/cert.pem && ( openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted2.de redacted2.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
1
```
==> Wouldn't run

**This change:**

* Cert doesn't exist:
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted.de/cert.pem
ls: cannot access /etc/letsencrypt/live/www.redacted.de/cert.pem: No such file or directory
root@lb:/# test -f /etc/letsencrypt/live/www.redacted.de/cert.pem && ! ( openssl x509 -in /etc/letsencrypt/live/www.redacted.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted.de redacted.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
1
```
==> Would run

* Cert does exists but one domain is missing
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted2.de/cert.pem
lrwxrwxrwx 1 root root 47 Dec 13 12:17 /etc/letsencrypt/live/www.redacted2.de/cert.pem -> ../../archive/www.redacted2.de/cert1.pem
root@lb:/# openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*'
DNS:redacted2.de
DNS:www.redacted2.de
root@lb:/# test -f /etc/letsencrypt/live/www.redacted2.de/cert.pem && ! ( openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted2.de redacted2.de foo.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
1
```
==> Would run

* Cert does exists and all domains match
```
root@lb:/# ls -la /etc/letsencrypt/live/www.redacted2.de/cert.pem
lrwxrwxrwx 1 root root 47 Dec 13 12:17 /etc/letsencrypt/live/www.redacted2.de/cert.pem -> ../../archive/www.redacted2.de/cert1.pem
root@lb:/# openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*'
DNS:redacted2.de
DNS:www.redacted2.de
root@lb:/# test -f /etc/letsencrypt/live/www.redacted2.de/cert.pem && ! ( openssl x509 -in /etc/letsencrypt/live/www.redacted2.de/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'www.redacted2.de redacted2.de' | tr ' ' '\n') | sort | uniq -c | grep -qv '^[ ]*2[ ]'
root@lb:/# echo $?
0
```
==> Wouldn't run

### DIFF
--- a/files/domain-validation.sh
+++ b/files/domain-validation.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+# THIS FILE IS MANAGED BY PUPPET
+
+# First parameter is the certificate path
+# All other parameters are the domain names requested for the certificate
+# Split whitespace into newlines to let it play nicely with sort
+certpath="${1}"
+shift
+domains=$(echo "${@}" | tr ' ' '\n')
+
+# If certpath doesn't exist, run the exec
+if ! test -f ${certpath}; then
+  exit 1
+fi
+
+# Get all subject alternative names from the certificate
+certdomains=$(openssl x509 -in ${certpath} -text -noout | grep -oP 'DNS:[^\s,]*' | sed 's/^DNS://g;')
+
+# Sort and uniq all domains. Drop all Domains which occure twice
+result=$(echo "${certdomains}\n${domains}" | sort | uniq -c | grep -Pcv '^[ \t]*2[ \t]')
+
+# If all requested domains are already in the certificate, the $result will be 0 otherwise > 0
+if [ ${result} -eq 0 ]; then
+  exit 0
+else
+  exit 1
+fi
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -97,9 +97,12 @@ define letsencrypt::certonly (
     command     => $command,
     path        => $::path,
     environment => $execution_environment,
-    unless      => "test -f ${live_path} && ! ( openssl x509 -in ${live_path} -text -noout | grep -oE 'DNS:[^\s,]*' | sed 's/^DNS://g;'; echo '${verify_domains}' | tr ' ' '\\n') | sort | uniq -c | grep -qv '^[ \t]*2[ \t]'",
+    unless      => "/usr/local/sbin/letsencrypt-domain-validation ${live_path} ${verify_domains}",
     provider    => 'shell',
-    require     => Class['letsencrypt'],
+    require     => [
+      Class['letsencrypt'],
+      File['/usr/local/sbin/letsencrypt-domain-validation'],
+    ],
   }
 
   if $ensure_cron  == 'present' {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -101,4 +101,13 @@ class letsencrypt (
     environment => concat([ "VENV_PATH=${venv_path}" ], $environment),
     refreshonly => true,
   }
+
+  # Used in letsencrypt::certonly Exec["letsencrypt certonly ${title}"]
+  file { '/usr/local/sbin/letsencrypt-domain-validation':
+    ensure => file,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0500',
+    source => "puppet:///modules/${module_name}/domain-validation.sh",
+  }
 }
--- a/spec/classes/letsencrypt_spec.rb
+++ b/spec/classes/letsencrypt_spec.rb
@@ -17,6 +17,15 @@ describe 'letsencrypt' do
 
           epel = facts[:osfamily] == 'RedHat'
 
+          it 'contains File[/usr/local/sbin/letsencrypt-domain-validation]' do
+            is_expected.to contain_file('/usr/local/sbin/letsencrypt-domain-validation').
+              with_ensure('file').
+              with_owner('root').
+              with_group('root').
+              with_mode('0500').
+              with_source('puppet:///modules/letsencrypt/domain-validation.sh')
+          end
+
           it 'contains the correct resources' do
             is_expected.to contain_class('letsencrypt::install').with(configure_epel: epel,
                                                                       manage_install: true,
--- a/spec/defines/letsencrypt_certonly_spec.rb
+++ b/spec/defines/letsencrypt_certonly_spec.rb
@@ -38,7 +38,7 @@ describe 'letsencrypt::certonly' do
         end
         it { is_expected.to contain_exec('initialize letsencrypt') }
         it { is_expected.to contain_exec('letsencrypt certonly foo.example.com') }
-        it { is_expected.to contain_exec('letsencrypt certonly foo.example.com').with_unless "test -f #{pathprefix}/etc/letsencrypt/live/foo.example.com/cert.pem && ! ( openssl x509 -in #{pathprefix}/etc/letsencrypt/live/foo.example.com/cert.pem -text -noout | grep -oE 'DNS:[^ ,]*' | sed 's/^DNS://g;'; echo 'foo.example.com' | tr ' ' '\\n') | sort | uniq -c | grep -qv '^[ \t]*2[ \t]'" }
+        it { is_expected.to contain_exec('letsencrypt certonly foo.example.com').with_unless "/usr/local/sbin/letsencrypt-domain-validation #{pathprefix}/etc/letsencrypt/live/foo.example.com/cert.pem foo.example.com" }
       end
 
       context 'with multiple domains' do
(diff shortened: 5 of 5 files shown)


baurmatt:
I'm pretty sure the failed test isn't related to my change? Do we have to do a modulesync run before?

bastelfreak:
@baurmatt can you please rebase against latest master?

baurmatt:
@bastelfreak Done! :)

baurmatt:
Hmm not sure what broke the Beaker tests -.-

Dan33l:
@baurmatt i will have a look about beaker tests to fix that

baurmatt:
Thanks! :) Done.

baurmatt:
Hmm yeah I know what you mean. I didn't put it in a script because i did not want to deploy a separate check script only for this exec. But if you want, I'm happy to rewrite it :)

Dan33l:
If you can, IMO the module will be better.

baurmatt:
Implemented the check script. Interesting thing about it: If you run Puppet in noop mode, it throws the following "error" if the check script isn't already implemented:

```
Letsencrypt::Certonly[foo.de]/Exec[letsencrypt certonly foo.de]: Could not evaluate: /bin/sh: 1: /usr/local/sbin/letsencrypt-domain-validation: not found
```

manifests/certonly.pp
```diff
@@ -97,7 +97,7 @@
     command     => $command,
     path        => $::path,
     environment => $execution_environment,
-    onlyif      => "test -f ${live_path} && ( openssl x509 -in ${live_path} -text -noout | grep -oE 'DNS:[^\s,]*' | sed 's/^DNS://g;'; echo '${verify_domains}' | tr ' ' '\\n') | sort | uniq -c | grep -qv '^[ \t]*2[ \t]'",
+    unless      => "/usr/local/sbin/letsencrypt-domain-validation ${live_path} ${verify_domains}",
```
Dan33l: Why not a require to `File['/usr/local/sbin/letsencrypt-domain-validation']` ?

manifests/init.pp
```diff
@@ -101,4 +101,13 @@
     environment => concat([ "VENV_PATH=${venv_path}" ], $environment),
     refreshonly => true,
   }
+
+  # Used in letsencrypt::certonly Exec["letsencrypt certonly ${title}"]
+  file { '/usr/local/sbin/letsencrypt-domain-validation':
```
Dan33l: The script is only used into `letsencrypt::certonly`, why is it created here and not with the same resource that need it ?

manifests/certonly.pp
baurmatt: That's a good idea! :)

manifests/init.pp
baurmatt: I've decided against moving it into `letsencrypt::certonly` because i would have to use `defined()` which might lead to unnecessary performance degradation on systems with a lot of `letsencrypt::certonly`'s. Also `letsencrypt::certonly` requires `Class[letsencrypt]` to be used, so this should work flawlessly.

baurmatt:
Not sure why some beaker tests fail while installing the Puppet Agent oO

Dan33l:
> Not sure why some beaker tests fail while installing the Puppet Agent oO

They are all related to `puppet6-nightly`. I think something is broken in nightly deb/rpm repository.

baurmatt:
Ha, I didn't get link. Thanks! :) Perhaps we should allow them to fail due to the nature of nightlies ;)

Dan33l:
I created a jira ticket about this:
https://tickets.puppetlabs.com/browse/PA-2417

files/domain-validation.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+
```
bastelfreak: please add a comment like:
```
# THIS FILE IS MANAGED BY PUPPET
```

manifests/init.pp
bastelfreak: isn't `/usr/local/sbin/` and it should be in `usr/local/bin` instead?

manifests/init.pp
```diff
@@ -101,4 +101,13 @@
...
+    ensure => file,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0500',
```
bastelfreak: Any reason not to use 0755 or even 0777?

files/domain-validation.sh
baurmatt: Done. :)

manifests/init.pp
baurmatt: May I answer this with a counter question? :) Any reason not to use '0500'?

We usually do this for scripts which are only used in Puppet which I find a good practice given that no other user should use that script. Its basically a simple `assert_private()` for scripts :D

manifests/init.pp
baurmatt: Per definition, scripts which are root only/system administration script are placed in */sbin. Thats the reason I chose this location.

PR #165, opened 2019-01-29
Add ability to control keysize. Default to 4096.

treveradams:
This allows keysize to be specified. 2048 is not exactly recommended for many things.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -45,6 +45,7 @@ define letsencrypt::certonly (
   Letsencrypt::Plugin                       $plugin               = 'standalone',
   Array[Stdlib::Unixpath]                   $webroot_paths        = [],
   String[1]                                 $letsencrypt_command  = $letsencrypt::command,
+  Integer[2048]                             $key_size             = $letsencrypt::key_size,
   Array[String[1]]                          $additional_args      = [],
   Array[String[1]]                          $environment          = [],
   Enum['present','absent']                  $ensure_cron          = 'absent',
@@ -62,9 +63,9 @@ define letsencrypt::certonly (
   }
 
   if ($custom_plugin) {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} "
   } else {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly -a ${plugin} "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} -a ${plugin} "
   }
 
   if $plugin == 'webroot' {
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -72,6 +72,7 @@ class letsencrypt (
   Boolean $agree_tos                     = $letsencrypt::params::agree_tos,
   Boolean $unsafe_registration           = $letsencrypt::params::unsafe_registration,
   Stdlib::Unixpath $config_dir           = $letsencrypt::params::config_dir,
+  Integer[2048] $key_size                = 4096,
 ) inherits letsencrypt::params {
 
   if $manage_install {
--- a/spec/defines/letsencrypt_certonly_spec.rb
+++ b/spec/defines/letsencrypt_certonly_spec.rb
@@ -46,7 +46,7 @@ describe 'letsencrypt::certonly' do
         let(:params) { { domains: ['foo.example.com', 'bar.example.com', '*.example.com'] } }
 
         it { is_expected.to compile.with_all_deps }
-        it { is_expected.to contain_exec('letsencrypt certonly foo').with_command "#{binaryprefix}letsencrypt --text --agree-tos --non-interactive certonly -a standalone --cert-name foo -d foo.example.com -d bar.example.com -d *.example.com" }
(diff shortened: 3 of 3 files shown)


manifests/certonly.pp
```diff
@@ -45,6 +45,7 @@
   Letsencrypt::Plugin                       $plugin               = 'standalone',
   Array[Stdlib::Unixpath]                   $webroot_paths        = [],
   String[1]                                 $letsencrypt_command  = $letsencrypt::command,
+  Optional[Integer]                         $key_size             = $letsencrypt::params::key_size,
```
ekohl: Since you have a default this can simply be `Integer`. Probably good to use `Integer[2048]` since we really don't want a smaller key size. It's also better to use `$letsencrypt::key_size` so users can override the default.

manifests/certonly.pp
```diff
@@ -62,9 +63,9 @@
   }
 
   if ($custom_plugin) {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly -rsa-key-size ${key_size}"
```
ekohl: Is it really a single dash? I'd expect `--rsa-key-size`

manifests/init.pp
```diff
@@ -67,6 +67,7 @@
   Boolean $manage_config                 = $letsencrypt::params::manage_config,
   Boolean $manage_install                = $letsencrypt::params::manage_install,
   Boolean $manage_dependencies           = $letsencrypt::params::manage_dependencies,
+  Integer $key_size                      = $letsencrypt::params::key_size,
```
ekohl: Can also be `Integer[2048]`

manifests/params.pp
```diff
@@ -9,6 +9,7 @@
   $venv_path           = '/opt/letsencrypt/.venv' # virtualenv path for vcs-installed letsencrypt
   $repo                = 'https://github.com/certbot/certbot.git'
   $cron_scripts_path   = "${facts['puppet_vardir']}/letsencrypt" # path for renewal scripts called by cron
+  $key_size            = 4096
```
ekohl: While I personally prefer 4096, AFAIK LE still defaults to 2048.

manifests/certonly.pp
treveradams: Thank you. I am merging in my changes to the module. I missed the - on a copy and paste.

manifests/certonly.pp
treveradams: Done as requested.

manifests/params.pp
treveradams: Do we need to follow the same thing? Or shall the default be the higher?

ekohl: Not sure. @Dan33l thoughts?

treveradams: I believe it should be 4096. If people want less, they can override. Security should be default in my opinion.

manifests/certonly.pp
```diff
@@ -45,6 +45,7 @@
   Letsencrypt::Plugin                       $plugin               = 'standalone',
   Array[Stdlib::Unixpath]                   $webroot_paths        = [],
   String[1]                                 $letsencrypt_command  = $letsencrypt::command,
+  Integer[2048]                             $key_size             = $letsencrypt::key_size,
```
bastelfreak: should we enforce 2048, 4096 and 8192 as only allowed values?

bastelfreak:
HI @treveradams, thanks for the PR! Can you take a look at the failed travis jobs and the inline comments?

manifests/certonly.pp
ekohl: I considered that, but 3072 is also used by some. Not sure how much we should limit it here

bastelfreak: ah true, let's stick with 2048 as a lower limit.

manifests/params.pp
Dan33l: In general, i try to stick with managed product defaults.
But about security, if a more secure setting does not cause trouble, i am fine with this.

treveradams:
I haven't updated letsencrypt_spec.rb. I am not sure how to access the os name fact here. For Fedora 29, it should be the same as RedHat 7, BUT it should not include epel.

manifests/certonly.pp
```diff
@@ -62,9 +63,9 @@
   }
 
   if ($custom_plugin) {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} "
   } else {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly -a ${plugin} "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly -a ${plugin} --rsa-key-size ${key_size} "
```
ekohl: Nit: I'd prefer to swap `-a` and `--rsa-key-size` so the start of the line is exactly the same for both with and without a custom plugin.

treveradams: The last push passes travis ci and addresses your concern. Unless I have missed something or there is something else, please consider accepting the pull request.

manifests/params.pp
Dan33l: So i agree with using default to 4096.
https://github.com/certbot/certbot/issues/489

treveradams:
I can do this and a force update if you would like.

treveradams:
I can do this and a force update if you would like.

bastelfreak:
Thanks!

PR #169, opened 2019-01-30
Add support for dns-rfc2136 plugin

Rathios:
Adds support for issuing certificates with the Let's Encrypt dns-rfc2136 plugin. Tested on CentOS7.

Three minor bonuses:
- Fixes a small spec bug where `dns-cloudflare` was not being allowed as a value for the Letsencrypt::Plugin type.
- Cleans up some old Puppet 3 syntax
- README gets a header upgrade for better readability (and so usage of this plugin can be documented sanely)

The module already has a dependency on puppetlabs/inifile, but I've used an ini template for (imo) cleaner code. Something to evaluate as well.

n/a

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -68,18 +68,35 @@ define letsencrypt::certonly (
     $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} -a ${plugin} "
   }
 
-  if $plugin == 'webroot' {
-    $_command_domains = zip($domains, $webroot_paths).map |$domain| {
-      if $domain[1] {
-        "--webroot-path ${domain[1]} -d ${domain[0]}"
-      } else {
-        "-d ${domain[0]}"
+  case $plugin {
+
+    'webroot': {
+      $_command_domains = zip($domains, $webroot_paths).map |$domain| {
+        if $domain[1] {
+          "--webroot-path ${domain[1]} -d ${domain[0]}"
+        } else {
+          "-d ${domain[0]}"
+        }
       }
+      $command_domains = join([ "--cert-name ${title}", ] + $_command_domains, ' ')
     }
-    $command_domains = join([ "--cert-name ${title}", ] + $_command_domains, ' ')
-  } else {
-    $_command_domains = join($domains, ' -d ')
-    $command_domains  = "--cert-name ${title} -d ${_command_domains}"
+
+    'dns-rfc2136': {
+      require letsencrypt::plugin::dns_rfc2136
+      $dns_args = [
+        "--cert-name ${title} -d",
+        join($domains, ' -d '),
+        "--dns-rfc2136-credentials ${letsencrypt::plugin::dns_rfc2136::config_dir}/dns-rfc2136.ini",
+        "--dns-rfc2136-propagation-seconds ${letsencrypt::plugin::dns_rfc2136::propagation_seconds}",
+      ]
+      $command_domains = join($dns_args, ' ')
+    }
+
+    default: {
+      $_command_domains = join($domains, ' -d ')
+      $command_domains  = "--cert-name ${title} -d ${_command_domains}"
+    }
+
   }
 
   if empty($additional_args) {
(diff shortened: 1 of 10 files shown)


manifests/certonly.pp
```diff
@@ -67,18 +67,35 @@
...
       }
+      $command_domains = join([ "--cert-name ${title}", ] + $_command_domains, ' ')
     }
-    $command_domains = join([ "--cert-name ${title}", ] + $_command_domains, ' ')
-  } else {
-    $_command_domains = join($domains, ' -d ')
-    $command_domains  = "--cert-name ${title} -d ${_command_domains}"
+
+    'dns-rfc2136': {
+      require ::letsencrypt::plugin::dns_rfc2136
```
bastelfreak: please don't add leading `::` in front of class names or variables. That is considered legacy syntax. It was required for Puppet 3 support.

manifests/plugin/dns_rfc2136.pp
```diff
@@ -0,0 +1,59 @@
...
+# [*manage_package*]
+#   Manage the plugin package.
+# [*config_dir*]
+#   The path to the configuration directory.
+#
+class letsencrypt::plugin::dns_rfc2136 (
+  String $server,
+  String $key_name,
+  String $key_secret,
+  String $key_algorithm = $letsencrypt::params::dns_rfc2136_algorithm,
```
bastelfreak: please align all the `=` in this section and enforce the minimal string length for the parameters:
`String[1]`

manifests/plugin/dns_rfc2136.pp
```diff
@@ -0,0 +1,59 @@
...
+  String $port          = $letsencrypt::params::dns_rfc2136_port,
```
bastelfreak: We should probably use this datatype? https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/port.pp

manifests/plugin/dns_rfc2136.pp
```diff
@@ -0,0 +1,59 @@
...
+  Boolean $manage_package = $letsencrypt::params::dns_rfc2136_manage_package,
+  String $config_dir      = $letsencrypt::config_dir,
```
bastelfreak: I assume this is always an absolute path? https://github.com/puppetlabs/puppetlabs-stdlib/blob/master/types/absolutepath.pp

bastelfreak:
Hi @Rathios, thanks for the PR. IMO a template is fine if we manage the whole file. I assume it isn't populated by any package? Or does certbot ship it by default?

manifests/plugin/dns_rfc2136.pp
alexjfisher: And `Stdlib::Host` for `$server`?

manifests/certonly.pp
Rathios: Good to know. There are 3 usages of it at the bottom of the file not related to my code. Update those at the same time?

There's also a couple of `${::path}` usages for execs, but those seem to require it to avoid local scope. Curious if that is an acceptable exception?

manifests/plugin/dns_rfc2136.pp
Rathios: Integer for `$propagation_seconds` probably wouldn't hurt as well.

manifests/certonly.pp
Rathios: Hrm. Those should probably be `$facts['path']` instead.

bastelfreak: correct. For facts, `$facts['path']` should be used.

Rathios:
Should be all good now.

Rathios:
... definitely done now. Curse you, tiny polishable details.

bastelfreak:
@Rathios can you take a look at the failing spec tests?

Rathios:
Rebased master and added new switch `--rsa-key-size 4096` to spec test, introduced by https://github.com/voxpupuli/puppet-letsencrypt/pull/165. Looks OK now. Thanks for the heads up.

Rathios:
Thanks!

PR #170, opened 2019-01-30
Add Debian Stretch to list of supported versions

ekohl:


metadata.json
```diff
@@ -43,7 +43,8 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "8"
+        "8",
+	"9"
```
bastelfreak: can you update the alignment? Should we also run the acceptance tests on Debian8/9?

dhoppe: I also think that we should add acceptance tests for Debian 8/9 and we should take this opportunity to drop acceptance tests for Ubuntu 14.04.

PR #128, opened 2018-07-23
Fix live directory for wildcard certificates

DocEmmetBrown:
#### Pull Request (PR) description

When issuing wildcard certificates, the created 'live' directory omits the wildcard character : 
`*.example.com` certificate files are located in `/etc/letsencrypt/live/example.com/` directory. 
The exec relies on the correct path for `creates`, in order not to run every time.
This PR removes the `*.` in the domain name when present.

PR #171, opened 2019-01-30
Add support for wildcard certs

Rathios:
Adds support for issuing wildcard certificates. If the first domain starts with an asterisk, certbot strips it and the first dot from the actual file path.

I don't think there is anything useful that can be added to spec tests for this. Please correct me if I'm wrong.

There is a stale pull request open for the same thing: https://github.com/voxpupuli/puppet-letsencrypt/pull/128

Also fixes a couple things in `files/domain_validation.sh` (I forgot to put these in a separate commit - let me know if you want them separate):

- It attempts to echo a newline with `echo "${certdomains}\n${domains}"`, but this does not work as expected on Red Hat-based distros (and probably others) as you need the (non-POSIX) switch `-e`. I've replaced this with a printf instead, with future support for a wider range of OSes in mind.
- Fixes unquoted variable warnings detected by shellcheck.

n/a

### DIFF
--- a/files/domain-validation.sh
+++ b/files/domain-validation.sh
@@ -9,18 +9,18 @@ shift
 domains=$(echo "${@}" | tr ' ' '\n')
 
 # If certpath doesn't exist, run the exec
-if ! test -f ${certpath}; then
+if ! test -f "${certpath}"; then
   exit 1
 fi
 
 # Get all subject alternative names from the certificate
-certdomains=$(openssl x509 -in ${certpath} -text -noout | grep -oP 'DNS:[^\s,]*' | sed 's/^DNS://g;')
+certdomains=$(openssl x509 -in "${certpath}" -text -noout | grep -oP 'DNS:[^\s,]*' | sed 's/^DNS://g;')
 
 # Sort and uniq all domains. Drop all Domains which occure twice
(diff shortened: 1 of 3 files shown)


files/domain-validation.sh
```diff
@@ -4,23 +4,23 @@
 # First parameter is the certificate path
 # All other parameters are the domain names requested for the certificate
 # Split whitespace into newlines to let it play nicely with sort
-certpath="${1}"
+certpath="$1"
```
bastelfreak: why do you remove all the braces? They are fine and don't do any harm.

Rathios: Shellcheck wants the variables in quotes. "$var" is a lot easier to type than "${var}". I changed the rest of them to be consistent with the vars I added. I can reverse it if you like.

bastelfreak:
Hi @Rathios, can you please rebase?

files/domain-validation.sh
bastelfreak: I agree with the quotes, but the braces aren't related. Can you please readd them? That makes the diff easier to read.

Rathios: Sure. :) Done. Vars are now consistent the other way, -with- braces.

Rathios:
Rebased, should be OK now.

bastelfreak:
Thanks!

Issue #175, opened 2019-02-25
Option to clean up cronjobs for removed domains


cFire:
There is currently no way for the module to clean up cronjobs when a domain is no longer in use.

## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: any
- Ruby: any
- Distribution: any
- Module version: 3.0.1-rc1

## How to reproduce (e.g Puppet code you use)
- Provision some certificate using `letsencrypt::certonly` with the `ensure_cron` option enabled.
- This domain stops being available for any reason (domain expired, DNS changed etc.)
- Try to de-provision cronjobs for renewing the certificate by setting `ensure_cron` to disabled.
- Module still tries (and of course fails) to get a certificate for the domain.
- Removing cronjobs fails because of failed dependencies.

## Output log
```
Error: 'certbot --text --agree-tos --non-interactive certonly --rsa-key-size 4096 -a apache --cert-name LE_redacted.com -d redacted.com' returned 1 instead of one of [0]
Error: /Stage[main]/Profile::Lydia/Profile::Lydia::User[redacted]/Site_apache::Vhost::User[redacted.com]/Letsencrypt::Certonly[LE_redacted.com]/Exec[letsencrypt certonly LE_redacted.com]/returns: change from 'notrun' to ['0'] failed: 'certbot --text --agree-tos --non-interactive certonly --rsa-key-size 4096 -a apache --cert-name LE_redacted.com -d redacted.com' returned 1 instead of one of [0]
```

## Any additional information you'd like to impart
This is a feature request, not a bug as the module is intended to function as it is doing, but there currently seems to be no options to deal with this sort of cleanup.

Dan33l:
It looks we didn't release it but what do you think about `ensure_cron` attribute ?
https://github.com/voxpupuli/puppet-letsencrypt/blob/master/manifests/certonly.pp#L51

cFire:
That's only half the puzzle. I'm using the pre-release version from git but it will still try to `exec { "letsencrypt certonly ${title}":`, which is no good if the domain is gone.

Perhaps we can add a normal `ensure` attribute which just takes 'absent/present'? That way we have a neat way both to clean up old certs, and to clean up cron jobs and scripts if enabled.

Dan33l:
You are right.

Can you propose a PR for this enhancement ?

cFire:
[...]

PR #174, opened 2019-02-11
Add support for certbot renew and hooks

Rathios:
#### Pull Request (PR) description
- Adds support for running `certbot renew` with hooks in cron
- Adds support for hooks in `certbot::certonly` (these are saved in the renew configs by certbot and will also be run by certbot when running `certbot renew`)
- Cleans up argument variables a bit in `certbot::certonly`, as it seemed overdue. Names and variable handling was getting confusing.
- Adds three new types: `Letsencrypt::Cron::Hour` `Letsencrypt::Cron::Minute` `Letsencrypt::Cron::Monthday`

#### This Pull Request (PR) fixes the following issues
Fixes #56

PR #161, opened 2019-01-16
Certonly: Refactor manage_cron to ensure_cron with appropriate data type

craigwatson:
#### Pull Request (PR) description
Ensures cron is removed when `manage_cron` is `false`.

#### This Pull Request (PR) fixes the following issues
Fixes #139

PR #176, opened 2019-02-25
release 4.0.0

Dan33l:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,42 @@ All notable changes to this project will be documented in this file.
 Each new release typically also includes the latest modulesync defaults.
 These should not affect the functionality of the module.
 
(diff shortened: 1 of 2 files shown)


Dan33l:
postponed.
Wait how #175 is resolved.

Rathios:
Would love to get this PR in as well. It should be ready. https://github.com/voxpupuli/puppet-letsencrypt/pull/174

Dan33l:
Hi @Rathios.

I postponed the release because a backward incompatible change was introduced by PR  #161 . And so, it will produce an increase of the major version.
But, the PR #175 can introduce an other change that should permit to not be backward incompatible anymore. 

Without a backward incompatible change we can release often, versions numbers are cheap as @igalic  said.

PR #177, opened 2019-02-26
Rework letsencrypt::certonly for #175

cFire:
WIP: Need to test it and update documentation etc.

+ Added $ensure attribute
+ Renamed $ensure_cron back to $manage_cron
+ Returned $manage_cron back to a Boolean value
+ Made execute of "letsencrypt certonly ${title}" conditional to $ensure == 'present'
+ Added cleanup of directory for domain certs when $ensure == 'absent'
+ Used global $ensure to signal desired state of cronjobs

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -5,6 +5,11 @@
 #
 # === Parameters:
 #
+# [*ensure*]
+#   Intended state of the resource. Accepts either 'present' or 'absent'.
+#   Default: 'present'.
+#   Will remove certificates for specified domains if set to 'absent'. Will
+#   also remove cronjobs and renewal scripts if `manage_cron` is set to 'true'.
 # [*domains*]
 #   Namevar. An array of domains to include in the CSR.
 # [*custom_plugin*]
@@ -23,9 +28,8 @@
 #   `letsencrypt-auto` command.
 # [*environment*]
 #   An optional array of environment variables (in addition to VENV_PATH).
-# [*ensure_cron*]
-#   Intended state of the cron and helper script resources. Accepts either
-#   'present' or 'absent'. Default: 'absent'
+# [*manage_cron*]
+#   Boolean indicating whether or not to schedule cron job for renewal. Default: 'false'.
 #   Runs daily but only renews if near expiration, e.g. within 10 days.
 # [*cron_before_command*]
 #   String representation of a command that should be run before renewal command
@@ -40,6 +44,7 @@
 #   e.g. 0 or '00' or [0,30].  Default - seeded random minute.
 #
 define letsencrypt::certonly (
+  Enum['present','absent']                  $ensure               = 'present',
   Array[String[1]]                          $domains              = [$title],
   Boolean                                   $custom_plugin        = false,
   Letsencrypt::Plugin                       $plugin               = 'standalone',
@@ -48,7 +53,7 @@ define letsencrypt::certonly (
   Integer[2048]                             $key_size             = $letsencrypt::key_size,
   Array[String[1]]                          $additional_args      = [],
   Array[String[1]]                          $environment          = [],
-  Enum['present','absent']                  $ensure_cron          = 'absent',
+  Boolean                                   $manage_cron          = false,
   Boolean                                   $suppress_cron_output = false,
   Optional[String[1]]                       $cron_before_command  = undef,
   Optional[String[1]]                       $cron_success_command = undef,
@@ -62,10 +67,14 @@ define letsencrypt::certonly (
     fail("The 'webroot_paths' parameter must be specified when using the 'webroot' plugin")
   }
 
-  if ($custom_plugin) {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} "
+  if $ensure == 'present' {
+    if ($custom_plugin) {
+      $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} "
+    } else {
+      $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} -a ${plugin} "
+    }
   } else {
-    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} -a ${plugin} "
+    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive delete "
   }
 
   case $plugin {
@@ -93,10 +102,13 @@ define letsencrypt::certonly (
     }
 
     default: {
-      $_command_domains = join($domains, ' -d ')
-      $command_domains  = "--cert-name ${title} -d ${_command_domains}"
+      if $ensure == 'present' {
+        $_command_domains = join($domains, ' -d ')
+        $command_domains  = "--cert-name ${title} -d ${_command_domains}"
+      } else {
+        $command_domains = "--cert-name ${title}"
+      }
     }
-
   }
 
   if empty($additional_args) {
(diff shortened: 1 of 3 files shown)


Dan33l:
Tests were failing due to timeout. I relaunched them and now its ok.

Dan33l:
@cFire can you add `[WIP]` (Work In Progress) in the title and remove it when you have finished the work ?

cFire:
The reason I've removed the WIP from the title is because it's ready for review (as far as I'm concerned). Sorry I did not explicitly mention that.

manifests/certonly.pp
```diff
@@ -145,6 +160,7 @@
     }
   } else {
```
Dan33l: When `$manage_cron` is set to `false` this means that cron should not be managed.
And so `else` is a nonsense. 

IMO, the `if $manage_cron` block should be moved in the code that handle `$ensure` with two differents  `if $manage_cron` blocks. A first one that explain what to do when `$ensure` is `present` and one other when `$ensure` is `absent'.

manifests/certonly.pp
```diff
@@ -112,21 +117,31 @@
...
+      environment => $execution_environment,
+      unless      => "/usr/local/sbin/letsencrypt-domain-validation ${live_path} ${verify_domains}",
+      provider    => 'shell',
+      require     => [
+        Class['letsencrypt'],
+        File['/usr/local/sbin/letsencrypt-domain-validation'],
+      ],
+    }
+  } else {
+    file { "${config_dir}/live/${live_path_domain}/":
```
Rathios: Did some testing to see what this'd do out of curiosity/concern. I'd personally look at running an exec to have certbot itself delete the certificate as rm -rf'ing the directory isn't quite sufficient to clean it all up. There are still files left behind in /etc/letsencrypt/renewal and /etc/letsecrypt/archive.

```
/etc/letsencrypt/renewal/certone.test.rawr.no.conf
/etc/letsencrypt/archive/certone.test.rawr.no
```

Certbot is unhappy about them as well:

```
$ certbot certificates
...
Renewal configuration file /etc/letsencrypt/renewal/certone.test.rawr.no.conf produced an unexpected error: expected /etc/letsencrypt/live/certone.test.rawr.no/cert.pem to be a symlink. Skipping.
```

Creating a new certificate again with the same name without cleaning up these files makes it append a -0001 to it, leaving the old files behind and generating a certificate name the module can't clean up by itself.

What about something like this?
```puppet
exec { "letsencrypt cleanup ${title}":
  command => "${letsencrypt_command} delete --cert-name ${title}",
  path    => $facts['path'],
  onlyif  => "test -d '${config_dir}/live/${live_path_domain}'",
}
```

cFire: That's a much better idea. I'll look into implementing that over the coming week.

cFire:
While working on implementing the delete function I noticed this;
```
$live_path_domain = regsubst($domains[0], '^\*\.', '')
$live_path = "${config_dir}/live/${live_path_domain}/cert.pem"
```
I think this is wrong when using a custom name like `letsencrypt::certonly { "special_snowflake_${domain}": [...] }` and will cause puppet to try and renew all certs every run.

Can someone please confirm I'm not seeing things that aren't there?
I think it should be
```
$live_path = "${config_dir}/live/${name}/cert.pem"
```

manifests/certonly.pp
cFire: Good call. I'll be implementing this next.

cFire:
Pull request is ready for another round of reviews. All tests are passing and it seems to work well in my environments.

manifests/certonly.pp
```diff
@@ -106,27 +111,38 @@
...
+      'unless'  => "/usr/local/sbin/letsencrypt-domain-validation ${live_path} ${verify_domains}",
+    }
+  } else {
+    $exec_params = {
+      'command' => "${letsencrypt_command} --text --agree-tos --non-interactive delete --cert-name ${title}",
+      'onlyif'  => "/usr/local/sbin/letsencrypt-domain-validation ${live_path} ${verify_domains}",
+    }
+  }
+
   exec { "letsencrypt certonly ${title}":
```
Dan33l: We should be able to forge `$command` yet in a generic manner.

At line 70 :
https://github.com/voxpupuli/puppet-letsencrypt/blob/e9808ad59a585cea2fb67d72b774104cbc843ed2/manifests/certonly.pp#L70
```
case $ensure {
  'present': {
    if ($custom_plugin) {
      $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} "
    } else {
      $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive certonly --rsa-key-size ${key_size} -a ${plugin} "
    }
  } 
  'absent': {
    $command_start = "${letsencrypt_command} --text --agree-tos --non-interactive delete "
  } 
  default: {
    fail ('unexpected value of ensure')
  } 
}
```


default at line 100
https://github.com/voxpupuli/puppet-letsencrypt/blob/e9808ad59a585cea2fb67d72b774104cbc843ed2/manifests/certonly.pp#L100
```
default: {
  case $ensure {
    'present': {
      $_command_domains = join($domains, ' -d ')
      $command_domains  = "--cert-name ${title} -d ${_command_domains}"
    }
    'absent': {
      $command_domains = "--cert-name ${title}"
    }
    default: {
      fail ('unexpected value of ensure')
    }
  } 
} 
```

Rathios:
@cFire The reason `${title}` is not used directly in the file path is because certbot generates the file path based on the first domain in the certificate (from `$domains[0]`). You can set the resource title independently from this, so if you pass both domains and a title, the path will probably be wrong. Certbot will also drop the wildcard if the first domain has one when generating the file path, eg. `*.example.com` gets stored in `/etc/letsencrypt/live/example.com`, thus the regsubst() to match that behavior.

Rathios:
@cFire I'm sorry, you were right - the sneaky feeling got me testing and certbot does use `--cert-name` to generate the file path and not `$domains[0]` / the first -d. It's easy to assume either since in some scenarios `$title` and `$domains[0]` will have the same value. So something like this should sort it? (With a comment to save others the same mistake)
```
# certbot uses --cert-name to generate the file path
$live_path_certname = regsubst($title, '^\*\.', '')
$live_path = "${config_dir}/live/${live_path_certname}/cert.pem"
```

manifests/certonly.pp
cFire: Is there any preference for
```
case $ensure {
  'present': {}
  'absent': {}
  default: {}
}
```
over
```
case $ensure {
  'present': {}
  default: {}
}
```
?

Since $ensure is an Enum that may only be 'present' or 'absent' it theoretically could never hit the default clause.

cFire: Or a third option;
```
if $ensure == 'present' {} else {}
```

alexjfisher: I'd probably go with option 3.

cFire: Job done, and all the tests are still happy.

manifests/certonly.pp
```diff
@@ -106,27 +118,35 @@
     $command_end = join(['',] + $additional_args, ' ')
   }
 
-  $command = "${command_start}${command_domains}${command_end}"
-  $live_path_domain = regsubst($domains[0], '^\*\.', '')
-  $live_path = "${config_dir}/live/${live_path_domain}/cert.pem"
+  # certbot uses --cert-name to generate the file path
+  $live_path_certname = regsubst($title, '^\*\.', '')
```
Dan33l: Is it related to this PR ? Looks an other issue.

cFire: To be fair that is a pre-existing issue, but I needed to fix it to properly test the changes on my real-world setup regardless.

Dan33l: It will be hard to find this bugfix in the middle of this PR that is not related with. It will not be listed in the CHANGELOG.md. And so, we should create a dedicated PR about this.

Can you explain the problem about path in an issue, and create a PR ? We will merge the PR about path issue, then rebase this #175, and merge it.

cFire: Created PR #180 to fix the path issue.

Dan33l: #180 is now merged. You can rebase #177

cFire:
I'm mildly confused about what just happened, but besides ci needing to be re-triggered it seems to be in a correct state again.

Dan33l:
> I'm mildly confused about what just happened, but besides ci needing to be re-triggered it seems to be in a correct state again.

I relaunched the travis ci.

cFire:
Well, that somehow broke absolutely everything. Stand by...

cFire:
Fixed again. Used the wrong rebase strategy before.

Dan33l:
A backward incompatible change was introduced by PR #161 . This PR #175 introduce an other change that permits a come back of `manage_cron` but introduce a new parameter `ensure`.

And so this PR is also backward incompatible.

Dan33l:
@cFire thank you for the job.

PR #185, opened 2019-04-17
Certificate facts

minorOffense:
Add facts about installed certificates so puppet code can make decisions based on the status of certificates.

Came from a discussion in https://github.com/voxpupuli/puppet-letsencrypt/issues/66

We've actually had this code around for a while and have been using it in production for over a year. I forgot to open the pull request earlier...

Credit should be given as well to @cwells who contributed the code for this.


minorOffense:
Ok, all refactored and still working. Ready for review

dhoppe:
@minorOffense Thank you very much for your contribution. Do you mind updating the documentation and adding some tests?

minorOffense:
I don’t mind updating the docs but I’m not sure how to write tests for this since it scans the contents of a certificate. Does your test suite make cert requests already? Or is there a dummy cert that’s added to the test run?

dhoppe:
@minorOffense Looks like we only have unit tests, but do not do any acceptance tests. https://github.com/voxpupuli/puppet-letsencrypt/blob/master/spec/defines/letsencrypt_certonly_spec.rb

Dan33l:
@dhoppe  some basics acceptance tests are done. But we do not get certificate during tests.

dhoppe:
@Dan33l Thx for the feedback. @minorOffense Please update the docs and were are good to go. ;)

minorOffense:
Sorry I was busy last week. I'll write up some docs and show a sample output from facter in a few different cases (invalid cert, valid, multiple domain cert). Should be ready for the end of the week.

minorOffense:
Sorry I forgot to remove the WIP. But yeah I added docs to it. Let me know if that's sufficient.

lib/facter/letsencrypt_directory.rb
```diff
@@ -0,0 +1,27 @@
+require 'openssl'
+require 'pathname'
+
+Facter.add(:letsencrypt_directory) do
+  setcode do
```
Dan33l: since this is not expected to work with windows node, the fact should be confined to Linux kernel.

minorOffense:
> I am not a ruby dev but i added an in line comment.
> And please squash your commits.

I'll add a note in the fact. For the squash I figured whoever merges it just uses the "Squash and merge" button in github. But I can "pre-squash" things.

lib/facter/letsencrypt_directory.rb
ekohl: How about BSD kernels? I'd expect it to work there as well.

Dan33l: Ah yes ! The module have a huge list of compatibles OSes.
This should work:
`confine :kernel => [:Linux, :FreeBSD, :OpenBSD]`

What i wanted to escape is to run the code of fact on windows.

minorOffense: Added the confine. And matched up to whatever rubocop wanted.

minorOffense: Whoops I did not squash that right...

minorOffense:
Ah I really messed up this branch. Ran one wrong git command and I've been digging a deeper hole ever since. I'm just going to reopen a fresh PR on a branch with everything correctly set.

minorOffense:
Ok I've fixed it all up and it's in https://github.com/voxpupuli/puppet-letsencrypt/pull/187

PR #187, opened 2019-05-17
feat(facts): add facts about certificates

minorOffense:
Add facts about installed certificates so puppet code can make decisions based on the status of certificates.

Came from a discussion in https://github.com/voxpupuli/puppet-letsencrypt/issues/66

We've actually had this code around for a while and have been using it in production for over a year. I forgot to open the pull request earlier...

Credit should be given as well to @cwells who contributed the code for this.

### DIFF
--- a/lib/puppet/functions/letsencrypt_lookup.rb
+++ b/lib/puppet/functions/letsencrypt_lookup.rb
@@ -1,0 +1,8 @@
+Puppet::Functions.create_function(:letsencrypt_lookup) do
+  def letsencrypt_lookup(cn)
+    domain = cn.split('.', 2)[1]
+    wildcard = "*.#{domain}"
+    certs = closure_scope['facts'].fetch('letsencrypt_directory', nil)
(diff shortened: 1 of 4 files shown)


lib/puppet/functions/letsencrypt_lookup.rb
```diff
@@ -0,0 +1,8 @@
+Puppet::Functions.create_function(:letsencrypt_lookup) do
```
bastelfreak: can you please namespace the function like `:'letsencrypt::lookup'`? The file needs to be moved to a `letsencrypt` subdirectory for this. Topscope functions are deprecated and we would like to not introduce more of them

minorOffense: I mirrored what they did in puppetlabs/apache https://github.com/puppetlabs/puppetlabs-apache/blob/master/lib/puppet/functions/apache/apache_pw_hash.rb

bastelfreak:
Thanks!

PR #189, opened 2019-06-27
Added support for cron environment in renew class

johansoetewey:
I added support to configure environments for the letsencrypt renew cron job. This way it possible to add for example an email address to send the output from the cronjob.


kenyon:
Closing in favor of #288.

PR #190, opened 2019-07-17
add debian9 as supported OS, enable acceptance tests with debian8 and debian9

markgraf:
added Debian 9, hope the spaceing is right :-)

https://github.com/voxpupuli/puppet-letsencrypt/pull/170 seems to be inactive since April.
Does this help?


metadata.json
```diff
@@ -43,7 +43,8 @@
     {
       "operatingsystem": "Debian",
       "operatingsystemrelease": [
-        "8"
+        "8",
+        "9"
```
Dan33l: Hi. @markgraf .
Can you also enable acceptance tests for debian ?

markgraf: Took me a while to smarten myself up a bit :-)
Next commit should do the trick.  I have no AWS-Account for testing though. :-/

markgraf:
`expected that the catalogue would contain Class[letsencrypt::install] with configure_epel set to true but it is set to false`

So I guess the acceptance-test needs some context?
configure_epel set to false sounds right for Debian :-)

Dan33l:
We use `beaker-hostgenerator` to generate nodesets on the fly.
So you have to update :
 * [.travis.yml](https://github.com/voxpupuli/puppet-letsencrypt/blob/master/.travis.yml) to enable tests for the PR
* [.sync.yml](https://github.com/voxpupuli/puppet-letsencrypt/blob/master/.sync.yml) to stay enabled after the next modulesync apply.

.sync.yml
```diff
@@ -5,5 +5,7 @@
     - set: ubuntu1404-64
     - set: ubuntu1604-64
     - set: ubuntu1804-64
+    - set: debianjessie-64
```
Dan33l: should be 
```suggestion
    - set: debian8-64
```

markgraf: Ninjaed :-)

spec/acceptance/nodesets/ec2/debian-jessie-x64.yml
```diff
@@ -0,0 +1,29 @@
+---
```
Dan33l: The file can be removed

spec/acceptance/nodesets/ec2/debian-stretch-x64.yml
```diff
@@ -0,0 +1,29 @@
+---
```
Dan33l: The file can be removed

spec/acceptance/nodesets/ec2/image_templates.yaml
```diff
@@ -27,6 +27,16 @@ AMI:
     :image:
       :aio: ami-fe408091
     :region: eu-central-1
+  # Debian 8.7 (HVM),
```
Dan33l: This modification can be removed

markgraf: Line 30-34, ok.
Lines 35-39 too?

spec/acceptance/nodesets/ec2/debian-jessie-x64.yml
markgraf: Ok, will do.

spec/acceptance/nodesets/ec2/debian-stretch-x64.yml
markgraf: Ok, will do.

spec/acceptance/nodesets/ec2/image_templates.yaml
Dan33l: please do not modify this file with the PR.

Dan33l:
Accordingly with `README.md` and [https://packages.debian.org](https://packages.debian.org/search?suite=jessie&searchon=names&keywords=letsencrypt) the package was not provided with debian8.

So, with this OS only `install_method=vcs` is possible. That explain the acceptance fail for debian8. We need to update the acceptance with debian8.

markgraf:
Debian8 can not find certbot-package.
Still failing because epel is expected. But `epel = facts[:osfamily] == 'RedHat'` and `configure_epel: epel,` should set this expectation to false, shouldn't it???

markgraf:
> the package was not provided with debian8.
>
> So, with this OS only `install_method=vcs` is possible.

That would be in spec/classes/letsencrypt_spec.rb lines 73 and following?

ekohl:
I also started https://github.com/voxpupuli/puppet-letsencrypt/pull/193 to clean up some things.

ekohl:
#193 has been merged instead.

markgraf:
Ok, I'll have to do manual installs then.  My company still uses Debian 8.  Yes, I know...

ekohl:
You can still override the install method on Debian Jessie to set it to VCS and I'd expect that to work.

PR #193, opened 2019-07-27
Drop Ubuntu 14.04 & add Debian 9/10 / Fedora 29/30 support

ekohl:
Adds Debian 9, Debian 10 and Fedora 30. Drops Debian 8 and Ubuntu 14.04.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,21 +24,6 @@ The module can integrate with [stahnma/epel](https://forge.puppetlabs.com/stahnm
 to set up the repo by setting the `configure_epel` parameter to `true` (the default for RedHat) and
 installing the module.
 
-On Debian Jessie the module assumes the package certbot is available. This
-package can be found in jessie-backports. When using
-[puppetlabs/apt](https://forge.puppet.com/puppetlabs/apt) the following code
-can be used:
-
-```puppet
-include apt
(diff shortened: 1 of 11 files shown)


bastelfreak:
Personally, I would be fine with keeping operating systems that still have upstream support in the metadata.json file.

ekohl:
When Debian 10 was released, the Debian 8 backports repositories were dropped. https://backports.debian.org/ states:
> Please keep in mind that backports doesn't follow LTS. Which means that we will drop support for oldstable (stretch) around one year after the release of buster. Thats in sync with the - official - security support for oldstable

We could set the type to VCS for that, but I'd prefer not to and consider it EOL.

bastelfreak:
@ekohl okay sounds good

README.md
```diff
@@ -171,7 +156,7 @@ Example:
 
 ```puppet
 class { 'letsencrypt::plugin::dns_rfc2136':
-  server     => '1.2.3.4',
+  server     => '192.0.2.1',
```
ekohl: This is TEST-NET-1 (192.0.2.0/24) from [rfc5737](https://datatracker.ietf.org/doc/rfc5737/).

PR #195, opened 2019-08-08
Add certbot-dns-ovh plugin support

aellert:
Add dns-ovh support based on dns_rfc2136 implementation

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -118,6 +118,17 @@ define letsencrypt::certonly (
       ]
     }
 
+    'dns-ovh': {
+      require letsencrypt::plugin::dns_ovh
+      $_domains = join($domains, '\' -d \'')
+      $plugin_args = [
+        "--cert-name '${title}' -d",
+        "'${_domains}'",
+        "--dns-ovh-credentials ${letsencrypt::plugin::dns_ovh::config_dir}/dns-ovh.ini",
+        "--dns-ovh-propagation-seconds ${letsencrypt::plugin::dns_ovh::propagation_seconds}",
+      ]
+    }
+
     default: {
       if $ensure == 'present' {
         $_domains = join($domains, '\' -d \'')
--- a/manifests/plugin/dns_ovh.pp
+++ b/manifests/plugin/dns_ovh.pp
@@ -1,0 +1,59 @@
+# == Class: letsencrypt::plugin::dns_ovh
+#
+#   This class installs and configures the Let's Encrypt dns-ovh plugin.
+#   https://certbot-dns-ovh.readthedocs.io
+#
+# === Parameters:
+#
+# [*endpoint*]
+#   Target OVH DNS endpoint.
+# [*application_key*]
+#   OVH application key.
+# [*application_secret*]
+#   DNS OVH application secret.
+# [*consumer_key*]
+#   DNS OVH consumer key.
+# [*manage_package*]
+#   Manage the plugin package.
+# [*package_name*]
+#   The name of the package to install when $manage_package is true.
+# [*config_dir*]
+#   The path to the configuration directory.
+#
+class letsencrypt::plugin::dns_ovh (
+  Enum['ovh-eu', 'ovh-ca'] $endpoint,
+  String[1] $application_key,
+  String[1] $application_secret,
+  String[1] $consumer_key,
+  Integer $propagation_seconds     = $letsencrypt::dns_ovh_propagation_seconds,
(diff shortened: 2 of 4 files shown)


manifests/certonly.pp
```diff
@@ -118,6 +118,17 @@
       ]
     }
 
+    'dns-ovh': {
+      require letsencrypt::plugin::dns_ovh
+      $_domains = join($domains, '\' -d \'')
+      $plugin_args = [
+        "--cert-name '${title}' -d",
+        "'${_domains}'",
+        "--dns-ovh-credentials ${letsencrypt::plugin::dns_ovh::config_dir}/dns-ovh.ini",
```
ekohl: Should the plugin declare a `$config_file` variable? Can be inside the body of the class. That way you don't rely on these two matching but can statically check it.

manifests/plugin/dns_ovh.pp
```diff
@@ -0,0 +1,59 @@
+# == Class: letsencrypt::plugin::dns_ovh
```
ekohl: Could you format this using puppet-strings style?

manifests/certonly.pp
aellert: I don't know exactly what you mean. Can you give me more details ?

manifests/plugin/dns_ovh.pp
aellert: Yes, it's done [here](https://github.com/aellert/puppet-letsencrypt/commit/1175fcc8073f2affb27a124367e017592c0aba83)
Should I do another PR ?

manifests/certonly.pp
ekohl: I meant something like (very shortened):
```puppet
class letsencrypt::plugin::dns_ovh {
  $config_file = "${letsencrypt::plugin::dns_ovh::config_dir}/dns-ovh.ini"

  file { $config_file:
    # ...
  }
}
```

Then you can use it here:
```puppet
"--dns-ovh-credentials ${letsencrypt::plugin::dns_ovh::config_file}",
```

manifests/plugin/dns_ovh.pp
ekohl: It's fine here. A separate PR to convert the rest of the module would be appreciated though :)

manifests/certonly.pp
aellert: Ok, I changed $config_dir by $config_file in this PR

manifests/plugin/dns_ovh.pp
aellert: Ok, next week I will probably have some time to do it !

aellert:
> It looks correct. Could you have a look at adding tests to this as well? Other than that 👍 to merging.

I never write tests before but I'm gonna try.

ekohl:
Any luck so far? I'd happy to make suggestions when you're stuck.

aellert:
Hi Ewoud,
I was busy last days..
But I managed to  add tests (largely inspired by rfc2136 plugin) and update README.
Note that travis failed because package python3-certbot-dns-ovh is only available from Debian 10.
How can we handle this ?

manifests/plugin/dns_ovh.pp
```diff
@@ -0,0 +1,87 @@
...
+#    consumer_key       => 'MDAwMDAwMDAwMDAwMDAwMDAwMDAwMDAw',
+#  }
+#  letsencrypt::certonly { 'foo':
+#    domains       => ['foo.example.com', 'bar.example.com'],
+#    plugin        => 'dns-ovh',
+#  }
+#
+# @see https://certbot-dns-ovh.readthedocs.io
+#
+# === Parameters:
```
ekohl: This is no longer needed with puppet-strings.
```suggestion
```

README.md
```diff
@@ -166,6 +166,39 @@ letsencrypt::certonly { 'foo':
...
+need to pass `endpoint`, `application_key`, `application_secret` and
+`consumer_key`to the class `letsencrypt::plugin::dns_ovh`.
+Ideally keys and secret should be encrypted, eg. with eyaml if using Hiera.
+
+Plugin documentation and it's parameters can be found here:
+https://certbot-dns-ovh.readthedocs.io
+
+Note:
+
+For Debian based OS, this plugin is compatible from Debian 10
```
ekohl: Why is it not supported? If there are no packages, would it be better to set the package name to undef on those and add a check in the class that if `manage_package` is true, the package must be set?

PR #198, opened 2019-08-23
Add support for email argument.

treveradams:
This adds support for the email flag. I do not yet have the CI acceptance stuff integrated. I am not sure how to add support for this where it should be optional.


treveradams:
Right now it appears to be failing due to the adding of the email argument in certonly. I see where the config.pp is supposed to create a file with it in it, but it is not happening. My cli.ini is empty. Hence this patch.

Dan33l:
Why using the already existing parameter `$config` or `$email` does not permit to configure an email address ?

Dan33l:
Hum ... using `config` is tested during acceptance : https://github.com/voxpupuli/puppet-letsencrypt/blob/master/spec/acceptance/letsencrypt_spec.rb#L27

And it looks to works.

Dan33l:
Unable to reproduce. If you can gives more information feel free to reopen.

PR #199, opened 2019-08-23
Add support for systemd service and timer units to renew.

treveradams:
This also adds support for naming services to restart with systemd instead of long-handing it all.
It also adds support for managing firewalls using firewalld for systems that are not meant to be http/https accessible at all times.

These three are essentially the same support and it is difficult to break it up

Again, I would love help knowing quite how to create the CI acceptances for this.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -78,6 +78,10 @@ class letsencrypt (
   $renew_cron_hour                       = fqdn_rand(24),
   $renew_cron_minute                     = fqdn_rand(60, fqdn_rand_string(10)),
   $renew_cron_monthday                   = '*',
+  String[1] $renew_systemd_datespec      = 'daily',
+  Enum['cron', 'systemd'] $timer_type    = 'cron',
+  Boolean          $manage_firewalld     = false,
+  Array[String[1]] $manage_services      = [],
 ) {
   if $manage_install {
     contain letsencrypt::install # lint:ignore:relative_classname_inclusion
--- a/manifests/renew.pp
+++ b/manifests/renew.pp
@@ -36,6 +36,10 @@ class letsencrypt::renew (
   Letsencrypt::Cron::Hour              $cron_hour            = $letsencrypt::renew_cron_hour,
   Letsencrypt::Cron::Minute            $cron_minute          = $letsencrypt::renew_cron_minute,
   Letsencrypt::Cron::Monthday          $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  String[1]                            $renew_systemd_datespec = $letsencrypt::renew_systemd_datespec,
+  Enum['cron', 'systemd']              $timer_type           = $letsencrypt::timer_type,
+  $manage_services                                           = $letsencrypt::manage_services,
+  Boolean $manage_firewalld                                  = $letsencrypt::manage_firewalld,
 ) {
   # Directory used for Puppet-managed renewal hooks. Make sure old unmanaged
   # hooks in this directory are purged. Leave custom hooks in the default
@@ -76,12 +80,56 @@ class letsencrypt::renew (
   ]).filter | $arg | { $arg =~ NotUndef and $arg != [] }
   $command = join($_command, ' ')
 
-  cron { 'letsencrypt-renew':
-    ensure   => $cron_ensure,
-    command  => $command,
-    user     => 'root',
-    hour     => $cron_hour,
-    minute   => $cron_minute,
-    monthday => $cron_monthday,
+  case $timer_type {
+    'cron': {
+      cron { 'letsencrypt-renew':
+        ensure   => $cron_ensure,
+        command  => $command,
+        user     => 'root',
+        hour     => $cron_hour,
+        minute   => $cron_minute,
+        monthday => $cron_monthday,
+      }
+    }
+    'systemd': {
+      $systemd_ensure = $facts['service_provider'] ? {
+        'systemd' => present,
+        default   => absent,
+      }
+
+      systemd::unit_file { 'letsencrypt.timer':
+        ensure  => $systemd_ensure,
+        content => epp('letsencrypt/letsencrypt.timer.epp'),
+      }
+
+      systemd::unit_file { 'letsencrypt.service':
+        ensure  => $systemd_ensure,
+        content => epp('letsencrypt/letsencrypt.service.epp'),
+      }
+
+      if systemd_ensure {
+        $systemd_timer = 'running'
+      } else {
+        $systemd_timer = $systemd_ensure
+      }
+
+      service { 'letsencrypt.timer':
+        ensure    => $systemd_timer,
+        enable    => true,
+        hasstatus => true,
+      }
+
+      [
+        '/etc/systemd/system/timers.target.wants/certbot.timer',
+        '/etc/systemd/system/certbot.timer',
+        '/etc/systemd/system/certbot.service',
+      ].each |Stdlib::Absolutepath $file| {
+        file { $file:
+          ensure => absent,
+        }
+      }
+    }
+    default: {
+    }
   }
 }
--- a/spec/spec_helper_acceptance.rb
+++ b/spec/spec_helper_acceptance.rb
@@ -8,4 +8,5 @@ configure_beaker do |host|
   when 'RedHat'
     host.install_package('crontabs')
   end
+  install_module_from_forge('puppet-systemd', '>= 2.6.0')
 end
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ The module can integrate with [puppet/epel](https://forge.puppetlabs.com/puppet/
 to set up the repo by setting the `configure_epel` parameter to `true` (the default for RedHat) and
 installing the module.
 
+For systemd support, [puppet/systemd](https://forge.puppet.com/modules/puppet/systemd) is required.
+
 ## Usage
 
 ### Setting up the Let's Encrypt client
--- a/spec/classes/plugin/dns_rfc2136_spec.rb
+++ b/spec/classes/plugin/dns_rfc2136_spec.rb
@@ -17,7 +17,7 @@ describe 'letsencrypt::plugin::dns_rfc2136' do
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
-        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-30', 'Fedora-31'
+        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-32'
           'python3-certbot-dns-rfc2136'
         when 'RedHat-7', 'CentOS-7'
           'python2-certbot-dns-rfc2136'
--- a/spec/classes/plugin/dns_route53_spec.rb
+++ b/spec/classes/plugin/dns_route53_spec.rb
@@ -17,7 +17,7 @@ describe 'letsencrypt::plugin::dns_route53' do
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
-        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-30', 'Fedora-31'
+        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-32'
           'python3-certbot-dns-route53'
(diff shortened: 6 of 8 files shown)


treveradams:
This is failing because I do not know how to tell travis how to pull in camptocamp/systemd. There may be other problems. I would appreciate some help knowing how to do this as I am not finding it in the travis documentation.

manifests/init.pp
```diff
@@ -82,6 +82,10 @@
   $renew_cron_hour                       = $letsencrypt::params::renew_cron_hour,
   $renew_cron_minute                     = $letsencrypt::params::renew_cron_minute,
   $renew_cron_monthday                   = $letsencrypt::params::renew_cron_monthday,
+  $systemd_date_spec                     = $letsencrypt::params::systemd_date_spec,
+  Enum['cron', 'systemd'] $timer_type    = $letsencrypt::params::timer_type,
+  Boolean          $manage_firewall      = $letsencrypt::params::manage_firewall,
```
Dan33l: We can have several way to manage firewall settings.

The way choosen to implement is with `firewalld` service and this is RedHat like OSes specific. A possibility looks to be to rename the parameter `manage_firewalld`.

And also, what happens if i use cron as `timer_type` and enable firewalld management ?

manifests/renew.pp
```diff
@@ -45,6 +45,10 @@
   Letsencrypt::Cron::Hour              $cron_hour            = $letsencrypt::renew_cron_hour,
   Letsencrypt::Cron::Minute            $cron_minute          = $letsencrypt::renew_cron_minute,
   Letsencrypt::Cron::Monthday          $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  $systemd_date_spec                                         = $letsencrypt::systemd_date_spec,
```
Dan33l: Please add custom data types.

manifests/init.pp
```diff
@@ -82,6 +82,10 @@
   $renew_cron_hour                       = $letsencrypt::params::renew_cron_hour,
   $renew_cron_minute                     = $letsencrypt::params::renew_cron_minute,
   $renew_cron_monthday                   = $letsencrypt::params::renew_cron_monthday,
+  $systemd_date_spec                     = $letsencrypt::params::systemd_date_spec,
```
Dan33l: To stay with same naming method, like with cron, can you rename this parameter `renew_systemd_datespec` ?

manifests/params.pp
```diff
@@ -6,6 +6,10 @@
   $manage_config       = true
   $manage_install      = true
   $manage_dependencies = true
+  $manage_firewall     = false
+  $manage_services     = []
+  $systemd_date_spec   = 'daily'
+  $timer_type          = 'systemd'
```
Dan33l: default must be `cron` or this PR will be backward incompatible.

manifests/renew.pp
```diff
@@ -86,13 +90,37 @@
...
+    cron { 'letsencrypt-renew':
+      ensure   => $cron_ensure,
+      command  => $command,
+      user     => 'root',
+      hour     => $cron_hour,
+      minute   => $cron_minute,
+      monthday => $cron_monthday,
+    }
+  } elsif $timer_type == 'systemd' {
+    ::systemd::unit_file { 'letsencrypt.timer':
```
Dan33l: You can remove leading coloms.

manifests/renew.pp
```diff
@@ -86,13 +90,37 @@
...
+      content => template('letsencrypt/letsencrypt.timer.erb'),
+    }
+    ::systemd::unit_file { 'letsencrypt.service':
```
Dan33l: same comment here.

manifests/renew.pp
```diff
@@ -86,13 +90,37 @@
...
+      content => template('letsencrypt/letsencrypt.service.erb'),
+    }
+    service { 'letsencrypt.timer':
+      ensure    => running,
+      enable    => true,
+      hasstatus => true,
+      restart   => 'service letsencrypt.timer restart',
```
Dan33l: With systemd, the command `service` is a old fashion method. The command `systemctl` should be preferred (probably with path).

spec/spec_helper_acceptance.rb
```diff
@@ -2,6 +2,7 @@
 require 'beaker-puppet'
 require 'beaker/puppet_install_helper'
 require 'beaker/module_install_helper'
+require 'camptocamp/systemd'
```
Dan33l: You have to remove this line.
The module `camptocamp/systemd` have to be installed with `spec_helper_acceptance.rb` if required by acceptances tests.

manifests/init.pp
```diff
@@ -82,6 +82,10 @@
...
+  Enum['cron', 'systemd'] $timer_type    = $letsencrypt::params::timer_type,
+  Boolean          $manage_firewall      = $letsencrypt::params::manage_firewall,
+  Optional[Array[String[1]]]  $manage_services      = $letsencrypt::params::manage_services,
```
Dan33l: Can you explain what this parameter is expected to do?

manifests/init.pp
Dan33l: In first, renaming the parameter with `firewalld` (that is a specific word) instead of `firewall` (that is a generic word) should permits to know that module manage a software that is named Firewalld and not a firewall in general.

In second, Firewalld exist also with Debian like OSes, even if it is not installed by default. So it is probably expected to be installed on Debian like OSes when `manage_firewalld` is set to `true`.

In third, the module is declared compatible with BSD like OSes. I don't know if Firewalld is available with.

The firewall layer should probably not be managed in the module, it was not with cron.

manifests/renew.pp
Dan33l: `String` is it possible ? I never used this, so i don't know.

manifests/renew.pp
```diff
@@ -86,13 +90,37 @@
...
   $command = join($_command, ' ')
 
-  cron { 'letsencrypt-renew':
-    ensure   => $cron_ensure,
-    command  => $command,
-    user     => 'root',
-    hour     => $cron_hour,
-    minute   => $cron_minute,
-    monthday => $cron_monthday,
+  if timer_type == 'cron' {
```
Dan33l: Please use `case` statement instead of `if`, it is easier to read.

manifests/renew.pp
```diff
@@ -86,13 +90,37 @@
...
+    cron { 'letsencrypt-renew':
+      ensure   => $cron_ensure,
+      command  => $command,
+      user     => 'root',
+      hour     => $cron_hour,
+      minute   => $cron_minute,
+      monthday => $cron_monthday,
+    }
+  } elsif $timer_type == 'systemd' {
+    systemd::unit_file { 'letsencrypt.timer':
```
Dan33l: Before using Systemd, we should check that Systemd is the init system used on the node. For this, the module [camptocamp/systemd custom fact](https://github.com/camptocamp/puppet-systemd/blob/master/lib/facter/systemd.rb) can be used.

manifests/renew.pp
Dan33l: Please prefer a `case` statement instead of `if`.

manifests/init.pp
Dan33l: We already manage in post and pre hook . Not sure it is safe to have to maintain two differents methods in a same module to make the same task.

spec/spec_helper_acceptance.rb
Dan33l: Like here : https://github.com/voxpupuli/puppet-etherpad/blob/master/spec/spec_helper_acceptance.rb#L14

manifests/renew.pp
Dan33l: Hum ... but why is it necessary to specify the `restart` attribute ?

treveradams:
I haven't yet addressed what you said about my manage_services. I am trying to figure out how to have this discussion with you or just carry this is my own fork of this (not the entire patch, just a small amount of it). I believe I have addressed everything else. It looks like it is going to pass Travis CI now.

treveradams:
I did have a small mishap with git, so if it appears things went sideways, please point it out. Same if I missed things.

manifests/renew.pp
```diff
@@ -86,13 +90,52 @@
...
+        ensure   => $cron_ensure,
+        command  => $command,
+        user     => 'root',
+        hour     => $cron_hour,
+        minute   => $cron_minute,
+        monthday => $cron_monthday,
+      }
+    }
+    'systemd': {
+      if $facts['systemd'] {
```
Dan33l: The fact use should ne like this:
```suggestion
       if $facts['service_provider'] == 'systemd' {
```
In the `if` block, we do configuration stuffs. In the `else` block, i understand we are in a situation without systemd available and we are trying to use it. So `else` block should return error with `fail()`.

Dan33l:
> I haven't yet addressed what you said about my manage_services. I am trying to figure out how to have this discussion with you.
You can ask to VP memeber on #voxpupuli on IRC preferably, or Slack.

manifests/init.pp
Dan33l: The current method can handle a list of command for hooks. And i don't think using two different methods for a same mechanism is a good thing.

manifests/init.pp
```diff
@@ -82,6 +82,10 @@
   $renew_cron_hour                       = $letsencrypt::params::renew_cron_hour,
   $renew_cron_minute                     = $letsencrypt::params::renew_cron_minute,
   $renew_cron_monthday                   = $letsencrypt::params::renew_cron_monthday,
+  String[1] $renew_systemd_datespec      = $letsencrypt::params::renew_systemd_datespec,
+  Enum['cron', 'systemd'] $timer_type    = $letsencrypt::params::timer_type,
+  Boolean          $manage_firewalld     = $letsencrypt::params::manage_firewalld,
+  Optional[Array[String[1]]]  $manage_services = $letsencrypt::params::manage_services,
```
bastelfreak: this should default to an empty array and not be optional:
```suggestion
  Array[String[1]] $manage_services    = $letsencrypt::params::manage_services,
```

manifests/renew.pp
bastelfreak: my suggestion here is to use the selector pattern:
https://puppet.com/docs/puppet/latest/lang_conditional.html#selectors

manifests/renew.pp
```diff
@@ -86,13 +90,52 @@
...
+        $systemd_ensure = present
+      } else {
+        $systemd_ensure = absent
+      }
+      systemd::unit_file { 'letsencrypt.timer':
+        ensure  => $systemd_ensure,
+        content => template('letsencrypt/letsencrypt.timer.erb'),
```
bastelfreak: please use epp instead of erb for new templates.

templates/letsencrypt.service.erb
```diff
@@ -0,0 +1,10 @@
+[Unit]
```
bastelfreak: please add a comment at the top like `# THIS FILE IS MANAGED BY PUPPET`

templates/letsencrypt.timer.erb
```diff
@@ -0,0 +1,11 @@
+[Unit]
```
bastelfreak: please add a comment at the top like `# THIS FILE IS MANAGED BY PUPPET`

manifests/renew.pp
treveradams: I believe I have fixed this to match what both of you have requested.

manifests/init.pp
treveradams: Fixed as requested.

manifests/renew.pp
treveradams: I have done this. I do not know how to check for an empty array in epp. I have it working, but not sure it is correct.

templates/letsencrypt.service.erb
treveradams: Done.

templates/letsencrypt.timer.erb
treveradams: Done.

treveradams:
I have still left in the short cut for restarting services. I am still trying to figure out the best way to do this as I need/want this locally, but want to maintain minimum changes.

kenyon:
Rebased, but now I'm not sure this is really needed since the distribution package (at least on Debian) provides systemd units for certbot.service and certbot.timer.

manifests/renew.pp
```diff
@@ -76,12 +80,56 @@
...
+        content => epp('letsencrypt/letsencrypt.service.epp'),
+      }
+
+      if systemd_ensure {
+        $systemd_timer = 'running'
+      } else {
+        $systemd_timer = $systemd_ensure
+      }
+
+      service { 'letsencrypt.timer':
```
bastelfreak, line 116: with more modern version of our systemd module, you can combine `systemd::unit_file { 'letsencrypt.timer':`, `systemd::unit_file { 'letsencrypt.service':` and the service resource:
https://github.com/voxpupuli/puppet-systemd#timer-units

ekohl:
I agree with @kenyon that we should encourage distro renewal scripts (cron and/or timer). For the Red Hat OS family that's `certbot-renew.timer` and for Debian there's `certbot.timer`. We should also encourage the use of the native renewal hooks.

PR #202, opened 2019-09-20
Raise upper bound version of stdlib & vcsrepo

mfaure:

As new versions of `stdlib` and `vcsrepo` have been released, this PR updates `metadata.json` to their latest version

(Any feedback would be greatly appreciated :) )

(none)

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -70,7 +70,7 @@
   "dependencies": [
     {
       "name": "puppetlabs/stdlib",
-      "version_requirement": ">= 4.13.1 < 6.0.0"
+      "version_requirement": ">= 4.13.1 < 7.0.0"
     },
     {
       "name": "puppetlabs/inifile",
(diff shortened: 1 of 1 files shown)


Dan33l:
@mfaure did you tested in your environment that module works well with the proposed versions of modules ? We have only some basics acceptance tests.

mfaure:
@Dan33l Thanks for your feedback. Honestly I relied upon Travis. What else should be done ?

ekohl:
@Dan33l why is this breaking? raising the upper bound is something I personally do even within a patch release.

Dan33l:
> @Dan33l why is this breaking? raising the upper bound is something I personally do even within a patch release.

I added the tag because changing the major version. But i am wrong because we do not change the minimal required version. thank you for your reading. I remove the tag.

ekohl:
Thanks!

PR #216, opened 2019-12-31
Add Fedora 31, drop Fedora 29

ekohl:
Fedora 29 is EOL since November 26th. Since there's no actual code change, it'll still continue to work for users. To make that happen, the Hiera hierachy is changed to allow setting variables per-OS.


dhoppe:
Puppet does not provide any packages for Fedora 31.

ekohl:
Should we drop the acceptance tests for it or not list it as supported at all?

dhoppe:
Good question. Even if Puppet provides the required package, facterdb might be the next problem.

ekohl:
On the other hand, nothing has really changed between F30 and F31 so how much value do the unit tests with facterdb actually provide?

dhoppe:
I wonder how much sense it makes to test a rolling release. We do not do that with any other module.

ekohl:
Fedora is not a rolling release. It has a [very clear 6 month release schedule](https://fedoraproject.org/wiki/Releases). this isn't for Fedora Rawhide and it's not like Arch.

dhoppe:
Good to know, but the Facts and Puppet packages will still not be available on time in the future.

Rathios:
Puppet packages for Fedora 31 have been available a while now. I'm one of those peeps using this module on Fedora 31, so would love to see this merged.

ekohl:
Restarted the Fedora 31 builds

ekohl:
@dhoppe any last comments?

PR #219, opened 2020-03-06
allow to override --cert-name

saimonn:

Add the possibility to choose a `$cert_name` parameter different from `$title`.

When the `$cert_name` vary with the `$title`, this prevents to use two resources to manage the same certificate, as in the following example:

```puppet
      # First standalone certificate is needed to create the key/certs, _before_ we can start apache with ssl enabled.
      ::letsencrypt::certonly {"${le_domains[0]}-standalone":
          domains     => $le_domains,
          manage_cron => false,
      }
      # this is the long-term cert, managing the cron renewal task
      -> ::letsencrypt::certonly {"${le_domains[0]}-webroot":
          domains              => $le_domains,
          manage_cron          => true,
          cron_success_command => 'apachectl -t && apachectl graceful',
          plugin               => 'webroot',
          webroot_paths        => [
            "/var/www/vhosts/${name}/htdocs",
          ],
      }
```

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -48,6 +48,7 @@
 define letsencrypt::certonly (
   Enum['present','absent']                  $ensure               = 'present',
   Array[String[1]]                          $domains              = [$title],
+  Optional[String[1]]                       $cert_name            = $title,
   Boolean                                   $custom_plugin        = false,
   Letsencrypt::Plugin                       $plugin               = 'standalone',
   Array[Stdlib::Unixpath]                   $webroot_paths        = [],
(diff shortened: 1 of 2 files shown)


alexjfisher:
https://github.com/voxpupuli/puppet-epel/pull/116 should fix the build failures. I've restarted the build to see. Same with your other PR.

saimonn:
Thanks !
I've fixed a trailing space in my spec test, it should pass now.

manifests/certonly.pp
```diff
@@ -48,6 +48,7 @@
 define letsencrypt::certonly (
   Enum['present','absent']                  $ensure               = 'present',
   Array[String[1]]                          $domains              = [$title],
+  Optional[String[1]]                       $cert_name            = $title,
```
alexjfisher: `Optional`? I think this should just be `String[1]`.  There's no way to set it to `undef` and that wouldn't be valid.

PR #221, opened 2020-03-28
#218 Switch to puppet-epel

### LINKED ISSUE #218: update metadata.json
metadata.json should be updated to switch to puppet-epel and allow puppetlabs-inifile v5

kallies:
As mentioned in #218 `stahnma-epel` has been transfered to voxpupuli. 

I left out `puppetlabs-inifile` < 6.0.0 as 5.x of this module is not released yet (current version is 4.1.0).

Fixes #218

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -81,8 +81,8 @@
       "version_requirement": ">= 2.0.0 < 4.0.0"
     },
     {
-      "name": "stahnma/epel",
-      "version_requirement": ">= 1.0.0 < 2.0.0"
+      "name": "puppet/epel",
+      "version_requirement": ">= 3.0.1 < 4.0.0"
     }
   ]
 }
--- a/.fixtures.yml
+++ b/.fixtures.yml
@@ -1,6 +1,6 @@
 fixtures:
   repositories:
-    epel: 'https://github.com/stahnma/puppet-module-epel.git'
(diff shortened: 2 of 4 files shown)


metadata.json
```diff
@@ -81,8 +81,8 @@
       "version_requirement": ">= 2.0.0 < 4.0.0"
     },
     {
-      "name": "stahnma/epel",
-      "version_requirement": ">= 1.0.0 < 2.0.0"
+      "name": "puppet/epel",
+      "version_requirement": ">= 1.0.0 < 4.0.0"
```
alexjfisher: The minimum version should be `3.0.0`.  There are no earlier versions in the `puppet` namespace.

alexjfisher:
Looks `puppet-epel` needs fixing. See https://github.com/voxpupuli/puppet-epel/pull/116

metadata.json
alexjfisher: Earlier today, I released `puppet/epel` `3.0.1`, which should fix the build failures.  Could you actually set that as the minimum version?

alexjfisher:
@kallies Thanks!  I guess you noticed https://github.com/voxpupuli/puppet-letsencrypt/pull/222 and force pushed to trigger a new build. Everything passes now. :)

PR #225, opened 2020-05-15
add manifest to install dns-route53 plugin, along with tests

aripringle:
The dns-route53 plugin was functional, but didn't manage the installation of the package. This creates a manifest to install it, along with tests.

I did not confirm the exact availability of every distribution - I assumed it would be available in the same versions that the dns-rfc2136 plugin is, so I copied that configuration. I have successfully tested these changes in our Centos7 environment.

This supports the only configuration option specific to that plugin, `--dns-route53-propagation-seconds`: https://certbot-dns-route53.readthedocs.io/en/stable/


ekohl:
Thanks!

PR #231, opened 2020-08-19
Allow setting file group owner

arthurbarton:
#### Add new optional parameter to set the group ownership of shipped files

Some Operating Systems (a current OpenBSD release) don't have group 'root' - this PR allows setting the group  to something else.
Default of group:root has not changed

N/A

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -25,6 +25,7 @@
 # @param config A hash representation of the letsencrypt configuration file.
 # @param cron_scripts_path The path for renewal scripts called by cron
 # @param cron_owner_group Group owner of cron renew scripts.
+# @param root_file_owner_group Group owner for the shipped files
 # @param manage_config A feature flag to toggle the management of the letsencrypt configuration file.
 # @param manage_install A feature flag to toggle the management of the letsencrypt client installation.
 # @param manage_dependencies A feature flag to toggle the management of the letsencrypt dependencies.
@@ -72,6 +73,7 @@ class letsencrypt (
   Hash $config                           = { 'server' => 'https://acme-v02.api.letsencrypt.org/directory' },
   String $cron_scripts_path              = "${facts['puppet_vardir']}/letsencrypt",
   String $cron_owner_group               = 'root',
+  String $root_file_owner_group          = 'root',
   Boolean $manage_config                 = true,
   Boolean $manage_install                = true,
   Boolean $manage_dependencies           = true,
@@ -124,7 +126,7 @@ class letsencrypt (
   file { '/usr/local/sbin/letsencrypt-domain-validation':
     ensure => file,
     owner  => 'root',
-    group  => 'root',
+    group  => $root_file_owner_group,
     mode   => '0500',
     source => "puppet:///modules/${module_name}/domain-validation.sh",
   }
--- a/manifests/plugin/dns_rfc2136.pp
+++ b/manifests/plugin/dns_rfc2136.pp
@@ -12,6 +12,7 @@
 # @param manage_package Manage the plugin package.
 # @param package_name The name of the package to install when $manage_package is true.
 # @param config_dir The path to the configuration directory.
+# @param ini_file_owner_group Group owner of the generated ini file
 #
 class letsencrypt::plugin::dns_rfc2136 (
   Stdlib::Host $server,
@@ -23,6 +24,7 @@ class letsencrypt::plugin::dns_rfc2136 (
   Integer $propagation_seconds     = 10,
   Stdlib::Absolutepath $config_dir = $letsencrypt::config_dir,
   Boolean $manage_package          = true,
+  String $ini_file_owner_group     = $letsencrypt::root_file_owner_group,
 ) {
   require letsencrypt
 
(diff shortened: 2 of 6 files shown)


metadata.json
```diff
@@ -7,58 +7,35 @@
   "source": "https://github.com/voxpupuli/puppet-letsencrypt",
   "project_page": "https://github.com/voxpupuli/puppet-letsencrypt",
   "issues_url": "https://github.com/voxpupuli/puppet-letsencrypt/issues",
-  "tags": [
-    "letsencrypt",
-    "let's encrypt",
-    "certbot",
-    "acme"
-  ],
+  "tags": ["letsencrypt", "let's encrypt", "certbot", "acme"],
```
ekohl, line 10: Would you mind keeping the previous formatting? The multiple lines make it easier to add/remove entries and keep the diffs smaller.

manifests/init.pp
```diff
@@ -72,6 +73,7 @@
   Hash $config                           = { 'server' => 'https://acme-v02.api.letsencrypt.org/directory' },
   String $cron_scripts_path              = "${facts['puppet_vardir']}/letsencrypt",
   String $cron_owner_group               = 'root',
+  String $root_file_owner_group          = 'root',
```
ekohl, line 76: I think this should be correct for the supported OSes. So it should be set to wheel on the right BSDs. In that case it should not be in `init.pp` at all, but only in data. See `configure_epel`.

The alternative is to not use a name but rather ID 0. That is common between unices.

metadata.json
arthurbarton: Sure thing! I think vim+coc auto 'fixed' this for me.

manifests/init.pp
arthurbarton: I've set this in the data now, thanks

metadata.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "puppet-letsencrypt",
-  "version": "5.0.1-rc0",
+  "version": "6.0.1-rc0",
```
ekohl, line 3: I don't think this is a major version.

manifests/init.pp
bastelfreak: since empty strins aren't allowed, please update to:
```suggestion
  String[1] $root_file_owner_group       = 'root',
```

manifests/plugin/dns_rfc2136.pp
```diff
@@ -23,6 +24,7 @@
   Integer $propagation_seconds     = 10,
   Stdlib::Absolutepath $config_dir = $letsencrypt::config_dir,
   Boolean $manage_package          = true,
+  String $ini_file_owner_group     = $letsencrypt::root_file_owner_group,
```
bastelfreak, line 27:
```suggestion
  String[1] $ini_file_owner_group  = $letsencrypt::root_file_owner_group,
```

manifests/renew.pp
```diff
@@ -26,16 +26,19 @@
...
-  Letsencrypt::Cron::Monthday          $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  Variant[String[1], Array[String[1]]] $pre_hook_commands     = $letsencrypt::renew_pre_hook_commands,
+  Variant[String[1], Array[String[1]]] $post_hook_commands    = $letsencrypt::renew_post_hook_commands,
+  Variant[String[1], Array[String[1]]] $deploy_hook_commands  = $letsencrypt::renew_deploy_hook_commands,
+  Array[String[1]]                     $additional_args       = $letsencrypt::renew_additional_args,
+  Enum['present', 'absent']            $cron_ensure           = $letsencrypt::renew_cron_ensure,
+  Letsencrypt::Cron::Hour              $cron_hour             = $letsencrypt::renew_cron_hour,
+  Letsencrypt::Cron::Minute            $cron_minute           = $letsencrypt::renew_cron_minute,
+  Letsencrypt::Cron::Monthday          $cron_monthday         = $letsencrypt::renew_cron_monthday,
+  String                               $hook_file_owner_group = $letsencrypt::root_file_owner_group,
```
bastelfreak, line 41:
```suggestion
  String[1]                            $hook_file_owner_group = $letsencrypt::root_file_owner_group,
```

metadata.json
bastelfreak: yes. please don't update the version. we do this during the next release.

metadata.json
```diff
@@ -50,7 +50,8 @@
     {
       "operatingsystem": "OpenBSD",
       "operatingsystemrelease": [
-        "6.2"
+        "6.6",
```
bastelfreak, line 53: mhm, is this a backwards incompatible change if we drop 6.2?

igalic: i reckon that the latest supported versions are the last two, so it's more like a cleanup

ekohl: Indeed. I think it also didn't really break compatibility with 6.2. That's just an EOL OS.

igalic (changes requested):
my issue with this PR is that we're introducing 4(?) parameters for 1 setting that was previously barely variable, or parametrisable.

so how about we just call this `root_group`, and be done with it?

how big is the chances that people will want to change this from the root (wheel) group to something else?

ekohl:
Btw, an alternative is to use the gid (`0`) instead of name. The gid is constant between Linux/BSD and avoids the need for a parameter.

PR #241, opened 2021-03-21
Fix certbot VCS package installation

### LINKED ISSUE #240: certbot-auto no longer works on any OS
See https://community.letsencrypt.org/t/certbot-auto-no-longer-works-on-debian-based-systems/139702/6

certbot-auto is no longer being supported, so therefore refuses to update/install certbot. This removal has been rolled out in monthly stages over the last few months. Their recommendation is to now use `snap` as the installation method.

When vcs is used as the install method, the certbot-auto script is called which now fails. As all the per-OS CI tests use the vcs installation method, this has had the effect of breaking CI for any change.

CallumBanbery:
As referenced in #240 the official installation method was changed
in December 2020, deprecating the use of the `certbot-auto` script,
and recommending `snap` as the primary package source.

As `snap` isn't yet a builtin package provider in Puppet and I don't
fancy creating a dependency on a third party module, nor do I want
to add a load of `exec` commands for managing snap packages I chose
to fix the VCS install method, and since it's used in the CI there's
less to change.

Since we can no longer use the `certbot-auto` script we should
follow the setup instructions provided by Letsencrypt's development
guide https://certbot.eff.org/docs/contributing.html.  Their guide
simply states to call a python script which configures the virtual
environment, and since the script is gone we no longer need the
VENV_PATH environment variable.

As the command needs to create a python virtual environment we
need to add the required package to the VCS dependencies, and from
testing I found there was no need to call the initialisation command
when installing from a package source, so I've made it conditional
to the VCS installation.

I've also taken the opportunity to update the version of `certbot`
that's checked out during VCS installation since `0.39.0` is now
around 1.5 years old.  Version `1.7.0` is the latest version that
is supported by Python 3.5 which is the lowest Python3 version
that's installed by default via package managers.

Fixes #240

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -105,17 +105,22 @@ class letsencrypt (
 
   if $manage_config {
     contain letsencrypt::config # lint:ignore:relative_classname_inclusion
-    Class['letsencrypt::config'] -> Exec['initialize letsencrypt']
+
+    if $install_method == 'vcs' {
+      Class['letsencrypt::config'] -> Exec['initialize letsencrypt']
+    }
   }
 
   contain letsencrypt::renew
 
-  exec { 'initialize letsencrypt':
-    command     => 'python3 tools/venv3.py',
-    cwd         => $path,
-    path        => $facts['path'],
-    environment => $environment,
-    refreshonly => true,
+  if $install_method == 'vcs' {
+    exec { 'initialize letsencrypt':
+      command     => 'python3 tools/venv3.py',
+      cwd         => $path,
+      path        => $facts['path'],
+      environment => $environment,
+      refreshonly => true,
+    }
   }
 
   # Used in letsencrypt::certonly Exec["letsencrypt certonly ${title}"]
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -14,8 +14,7 @@
 #   The email address to use to register with Let's Encrypt. This takes
 #   precedence over an 'email' setting defined in $config.
 # @param path The path to the letsencrypt installation.
-# @param venv_path virtualenv path for vcs-installed Certbot
-# @param environment An optional array of environment variables (in addition to VENV_PATH)
+# @param environment An optional array of environment variables.
 # @param repo A Git URL to install the Let's encrypt client from.
 # @param version The Git ref (tag, sha, branch) to check out when installing the client with the `vcs` method.
 # @param package_name Name of package and command to use when installing the client with the `package` method.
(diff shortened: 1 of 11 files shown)


manifests/init.pp
```diff
@@ -75,6 +74,7 @@
   Boolean $manage_config                 = true,
   Boolean $manage_install                = true,
   Boolean $manage_dependencies           = true,
+  Array[String] $vcs_dependencies        = [],
```
bastelfreak, line 77: I think allowing empty strings doesn't make sense here so we can prohibit it?
```suggestion
  Array[String[1]] $vcs_dependencies     = [],
```

manifests/install.pp
```diff
@@ -20,12 +21,12 @@
   String $path                           = $letsencrypt::path,
   String $repo                           = $letsencrypt::repo,
   String $version                        = $letsencrypt::version,
+  Array[String] $vcs_dependencies        = $letsencrypt::vcs_dependencies,
```
bastelfreak, line 24: I think  empty are also not here allowed / don't make sense?
```suggestion
  Array[String[1]] $vcs_dependencies     = $letsencrypt::vcs_dependencies,
```

manifests/init.pp
```diff
@@ -112,11 +110,11 @@
 
   contain letsencrypt::renew
 
-  # TODO: do we need this command when installing from package?
   exec { 'initialize letsencrypt':
```
ekohl, line 113: I think this only works with the VCS checkout. Perhaps it should be hidden behind `if $install_method == 'vcs' {` as well?

manifests/init.pp
CallumBanbery: Updated

manifests/install.pp
CallumBanbery: Updated

manifests/init.pp
CallumBanbery: Updated

manifests/init.pp
```diff
@@ -91,33 +91,36 @@
...
 
-  # TODO: do we need this command when installing from package?
-  exec { 'initialize letsencrypt':
-    command     => "${command_init} -h",
-    path        => $facts['path'],
-    environment => concat(["VENV_PATH=${venv_path}"], $environment),
-    refreshonly => true,
+  if $install_method == 'vcs' {
+    exec { 'initialize letsencrypt':
+      command     => 'python3 tools/venv3.py',
```
ekohl, line 118: What does this do? Isn't this sort of an install command?

spec/classes/letsencrypt_install_spec.rb
```diff
@@ -21,6 +22,14 @@
       let :facts do
         facts
       end
+      let(:vcs_dependencies) do
+        case facts[:osfamily]
+        when 'Debian'
+          ['git', 'python3', 'python3-venv']
+        when 'RedHat', 'OpenBSD', 'FreeBSD'
```
ekohl, line 29: I don't see data files for the BSDs to set there properties. How are these set?

manifests/init.pp
bastelfreak: does that setup the virtualenv? Doesn't it make sense to use the voxpupuli/python module to create it?

spec/classes/letsencrypt_install_spec.rb
lschierer: I see that there are data/FreeBSD-family.yaml and data/OpenBSD-family.yaml files, but that they do not have letsencrypt::vcs_dependencies sections in them.  I do not know what the dependencies for letsencrypt would be on those systems to try adding them.  Is that what is blocking this PR?

Dan33l:
Since the VCS method is removed from the module, i close this.

PR #235, opened 2020-11-25
modulesync 4.1.0 / Drop EoL Puppet 5 / Drop VCS install method

bastelfreak:
modulesync 4.0.0

PR #244, opened 2021-04-27
support for Ubuntu 20.04

lschierer:
When I added a Ubuntu 20.04 node to my puppet master, it could not find the package name for the DNS plugin.  This fixed it.


kenyon:
We would also need to update metadata.json to declare support for Ubuntu 20.04.

lschierer:
I have updated that file now as well, I missed it before.

metadata.json
```diff
@@ -38,6 +38,7 @@
       "operatingsystemrelease": [
         "16.04",
         "18.04"
```
kenyon, line 40:
```suggestion
        "18.04",
```

kenyon (changes requested):

metadata.json
lschierer: done

kenyon (commented):
Thanks. Looks like the remaining work to get the tests to pass is in #235 and possibly other PRs.

lschierer:
can we remove the needs work label, or is there more I should do?

lschierer:
it looks like a maintainer has to approve the workflow now to see if the latest merge from upstream resolves any of the test failures.

lschierer:
it appears I am now passing all checks, so the tests-fail tag added by https://github.com/voxpupuli/puppet-letsencrypt/pull/244#issuecomment-858544263 should be removed.

ekohl (approved):

ekohl:
@kenyon I think your concerns have been addressed, but you are listed as "requested changes" now. Any objections to merging this?

Also, we should squash this. We can do that while merging, but IMHO it's better for a cleaner git log.

lschierer:
now squashed per request.

ekohl:
Looks like something went wrong there because I see 18 commits. Can you use `git rebase`?

lschierer:
I cannot figure out enough git rebase to clean this up. I'm going to close this and recreate it.

PR #256, opened 2021-11-15
dns_rfc2136_spec, dns_route53_spec: add Fedora 32, drop Fedora 30, 31

kenyon:
Support for Fedora 30 and 31 was removed in #235.

### DIFF
--- a/spec/classes/plugin/dns_rfc2136_spec.rb
+++ b/spec/classes/plugin/dns_rfc2136_spec.rb
@@ -17,7 +17,7 @@ describe 'letsencrypt::plugin::dns_rfc2136' do
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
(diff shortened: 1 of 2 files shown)


root-expert (approved):

spec/classes/plugin/dns_rfc2136_spec.rb
```diff
@@ -17,7 +17,7 @@ class { 'letsencrypt':
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
-        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-30', 'Fedora-31'
+        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-32'
```
ekohl, line 20: Can't we use `/^Fedora-/` instead to make it ready for the future? Same with Debian and Ubuntu.

PR #258, opened 2021-11-15
init.pp: remove unused param $manage_dependencies

kenyon:
Usage of this class parameter was removed in commit ee77abc6ea9ca114f8fec382522abfe56571daef.


kenyon:
Test failure is fixed in #256.

ekohl (approved):
Is this a breaking change? Technically it can break a catalog but it'd be a bit excessive to do a major bump just for this.

kenyon:
> Is this a breaking change? Technically it can break a catalog but it'd be a bit excessive to do a major bump just for this.

I don't think it can cause breakage because the parameter was only used for the VCS install method which you removed in commit ee77abc. So users would have already had to update their usage of this module to handle that change, if they were using the VCS install method.

PR #262, opened 2021-12-10
add puppetlabs-cron_core as dependency

Dan33l:
Since minimal version of Puppet is 6, the core modules have to be declared as a dependency

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -83,6 +83,10 @@
       "version_requirement": ">= 2.0.0 < 6.0.0"
     },
     {
+      "name": "puppetlabs-cron_core",
+      "version_requirement": ">= 1.0.0 < 2.0.0"
+    },
+    {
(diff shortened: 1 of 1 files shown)


alexjfisher:
But do they?  I thought our policy was not to (unless we require a specific version) since they're automatically vendored.

ekohl:
Indeed, it is expected that Puppet packages ensure `*_core` packages are present already so these shouldn't need to be declared. This can have ugly side effects, such as overriding it in an environment where it masks a newer version or pulls in a too new version for an environment. I don't think we should add these.

Dan33l:
I was not aware about a policy about *_core modules.  But the CI is green only because the core module is in `.fixtures.yml`. So why is it added ? Because tests fails without.

I don't find any login in this but Puppet Inc looks to not add _core modules too.

PR #264, opened 2021-12-14
depend on puppetlabs-cron_core

nod0n:
`puppetlabs-cron_core` is listed in the [`.fixtures.yml`](https://github.com/voxpupuli/puppet-letsencrypt/blob/master/.fixtures.yml#L9-L11) but not in the `metadata.json`. 
The `puppetlabs-cron_core` module is used in [`cron.pp`](https://github.com/voxpupuli/puppet-letsencrypt/blob/master/manifests/renew.pp#L79-L86).

All modules that this module depends on, should be listed in the `metadata.json`.

### Note
I don't know if the version I chose, is appropriate. Any suggestions?

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -85,6 +85,10 @@
     {
       "name": "puppet/epel",
       "version_requirement": ">= 3.0.1 < 5.0.0"
+    },
+    {
+      "name": "puppetlabs/cron_core",
+      "version_requirement": ">= 1.1.0 < 2.0.0"
     }
   ]
 }


bastelfreak:
Hi,
those *_core modules are bundled within the puppet agent, that's why they are not listed in the metadata.json. This is on purpose :) but moved the code from the actual puppet agent into modules to make their development easier, but for each release they are vendored into the agent again.

ekohl:
This is the second PR we get about this. Feels like we should have some document to refer to. Is there an official Puppet source?

nod0n:
Maybe just a little comment in the `metadata.json`(?)

ekohl:
No, JSON has no comments. It's also something we would need to do in every single module which is a lot of work. This should be in some reference documentation. Perhaps https://github.com/voxpupuli/metadata-json-lint could also warn about it.

PR #265, opened 2021-12-14
Drop obsolete operating system version check in acceptance tests

nod0n:
Debian 9 and Ubuntu 16.04 are not supported by this plugin, so therefore
this check is obsolete.
Fixes: https://github.com/voxpupuli/puppet-letsencrypt/commit/415b8827101d6e473d8fb667b0426ea2a8731608

### DIFF
--- a/spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
+++ b/spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
@@ -1,16 +1,6 @@
 require 'spec_helper_acceptance'
 
 describe 'letsencrypt::plugin::dns_rfc2136' do
-  supported = case fact('os.family')
-              when 'Debian'
-                # Debian 9 has it in backports, Ubuntu started shipping in Bionic
-                fact('os.release.major') != '9' && fact('os.release.major') != '16.04'
-              when 'RedHat'
-                true
-              else
-                false
-              end
-
   context 'with defaults values' do
     pp = <<-PUPPET
       class { 'letsencrypt' :
@@ -26,24 +16,18 @@ describe 'letsencrypt::plugin::dns_rfc2136' do
       }
     PUPPET
 
-    if supported
-      it 'installs letsencrypt and dns rfc2136 plugin without error' do
-        apply_manifest(pp, catch_failures: true)
-      end
-      it 'installs letsencrypt and dns rfc2136 idempotently' do
-        apply_manifest(pp, catch_changes: true)
-      end
+    it 'installs letsencrypt and dns rfc2136 plugin without error' do
+      apply_manifest(pp, catch_failures: true)
+    end
+    it 'installs letsencrypt and dns rfc2136 idempotently' do
+      apply_manifest(pp, catch_changes: true)
+    end
 
-      describe file('/etc/letsencrypt/dns-rfc2136.ini') do
-        it { is_expected.to be_file }
-        it { is_expected.to be_owned_by 'root' }
-        it { is_expected.to be_grouped_into 'root' }
-        it { is_expected.to be_mode 400 }
-      end
-    else
-      it 'fails to install' do
-        apply_manifest(pp, expect_failures: true)
-      end
+    describe file('/etc/letsencrypt/dns-rfc2136.ini') do
+      it { is_expected.to be_file }
+      it { is_expected.to be_owned_by 'root' }
+      it { is_expected.to be_grouped_into 'root' }
+      it { is_expected.to be_mode 400 }
     end
   end
 end
--- a/spec/acceptance/letsencrypt_plugin_dns_route53_spec.rb
+++ b/spec/acceptance/letsencrypt_plugin_dns_route53_spec.rb
@@ -1,16 +1,6 @@
 require 'spec_helper_acceptance'
 
 describe 'letsencrypt::plugin::dns_route53' do
-  supported = case fact('os.family')
-              when 'Debian'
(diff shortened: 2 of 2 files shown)


spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
```diff
@@ -3,8 +3,7 @@
 describe 'letsencrypt::plugin::dns_rfc2136' do
   supported = case fact('os.family')
```
ekohl, line 4: You could also write the whole definition now as:
```ruby
supported = ['Debian', 'RedHat'].include?(fact('os.family'))
```

You can even natively skip things:
```ruby
 describe 'letsencrypt::plugin::dns_rfc2136', if: ['Debian', 'RedHat'].include?(fact('os.family')) do
```

Then you can drop `if supported` on line 28` as well.

TuningYourCode (approved):

ekohl (approved):
I like to check which commit dropped it and then add `Fixes: COMMIT` to it. In this case it's 415b8827101d6e473d8fb667b0426ea2a8731608.

I also don't mind the current patch, but you could go further. See inline comments.

ekohl (approved):
I love patches/PRs with net-negative lines while achieving the same :)

Edit: looks like Rubocop prefers a different syntax, but that's a trivial change.

spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
nod0n: It didn't come to my mind, that a named parameter would be called `if`. The ruby irc helped me understand. :)

I tried to implement the if filter.

For me the question came up, is it not possible to support this two certbot plugins on the BSD systems supported by this module? This would improve the tests even more.

ekohl: Maybe? I don't know BSD. If there's a package then yes. We can also argue that we don't run acceptance tests on BSD at all now so the whole `if:` could be dropped. If we do add BSD, then we can deal with it then?

spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
```diff
@@ -1,16 +1,6 @@
...
-              when 'Debian'
-                # Debian 9 has it in backports, Ubuntu started shipping in Bionic
-                fact('os.release.major') != '9' && fact('os.release.major') != '16.04'
-              when 'RedHat'
-                true
-              else
-                false
-              end
-
+describe 'letsencrypt::plugin::dns_rfc2136', if: %w[Debian RedHat].include?(fact('os.family')) do
```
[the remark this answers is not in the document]
nod0n: I don't know. I just wanted to drop an obsolete check. Maybe we can support this two plugins for all supported operating systems and this would not be necessary.
@ekohl Do you have any inside?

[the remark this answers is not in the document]
ekohl: In https://github.com/voxpupuli/puppet-letsencrypt/pull/265#discussion_r768772759 I suggested the same. So I think it'd be fine to drop the whole condition. So
```suggestion
describe 'letsencrypt::plugin::dns_rfc2136' do
```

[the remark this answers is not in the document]
nod0n: I found packages for FreeBSD bot I could not find packages for OpenBSD. Shall I add them in a separate PR?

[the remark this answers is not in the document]
nod0n: I'll create a separate PR.

nod0n:
see also https://github.com/voxpupuli/puppet-letsencrypt/pull/268

spec/acceptance/letsencrypt_plugin_dns_route53_spec.rb
```diff
@@ -1,16 +1,6 @@
...
-              when 'Debian'
-                # Debian 9 has it in backports, Ubuntu started shipping in Bionic
-                fact('os.release.major') != '9' && fact('os.release.major') != '16.04'
-              when 'RedHat'
-                true
-              else
-                false
-              end
-
+describe 'letsencrypt::plugin::dns_route53', if: %w[Debian RedHat].include?(fact('os.family')) do
```
ekohl, line 3: You can drop this too
```suggestion
describe 'letsencrypt::plugin::dns_route53' do
```

spec/acceptance/letsencrypt_plugin_dns_rfc2136_spec.rb
```diff
@@ -26,24 +16,18 @@ class { 'letsencrypt::plugin::dns_rfc2136':
...
-      end
-      it 'installs letsencrypt and dns rfc2136 idempotently' do
-        apply_manifest(pp, catch_changes: true)
-      end
+    it 'installs letsencrypt and dns rfc2136 plugin without error' do
+      apply_manifest(pp, catch_failures: true)
+    end
+    it 'installs letsencrypt and dns rfc2136 idempotently' do
+      apply_manifest(pp, catch_changes: true)
+    end
```
ekohl, line 24: Not for this PR, but this can be implemented via [a shared example](https://github.com/voxpupuli/voxpupuli-acceptance/#an-idempotent-resource)

ekohl (commented):
Could you squash all commits?

ekohl (approved):
Thanks!

PR #197, opened 2019-08-19
Add Nginx authenticator plugin to Letsencrypt Puppet module

alexfouche:
#### Pull Request (PR) description

I added automatic installation of the `...-certbot-nginx-...` package when `'plugin' => 'nginx'` is given to define `letsencrypt::certonly`. And i provided an example in Readme.

#### This Pull Request (PR) fixes the following issues

On Centos for example, when a Nginx web server is running, simply asking Certbot to create or renew certificates is not enough. It requires an additional package or plugin,

PR #267, opened 2021-12-14
install and use the certbot nginx plugin

nod0n:
A try to get the Nginx authenticator in this module. I started from https://github.com/voxpupuli/puppet-letsencrypt/pull/197, tried to throw out what was not related to the nginx plugin and tried to simplify where possible.
The nginx certbot plugin has currently to be installed outside of this module.


data/os/CentOS/7.yaml
```diff
@@ -1,3 +1,4 @@
 ---
-letsencrypt::plugin::dns_rfc2136::package_name: 'python2-certbot-dns-rfc2136'
-letsencrypt::plugin::dns_route53::package_name: 'python2-certbot-dns-route53'
+letsencrypt::plugin::dns_rfc2136::package_name: "python2-certbot-dns-rfc2136"
```
bastelfreak, line 2: why did you switch to double quotes? single quotes should be fine

nod0n: Single quotes are fine. My editor just prefers them for some reason (it replaces them on every save operation).

nod0n: I changed my editor's behavior and force pushed.

bastelfreak (approved):

README.md
```diff
@@ -120,6 +120,18 @@ letsencrypt::certonly { 'foo':
...
+
+To request a certificate for `foo.example.com` and `bar.example.com` with the
+`certonly` installer and the `nginx` authenticator:
+
+```puppet
+letsencrypt::certonly { 'foo':
+  domains => ['foo.example.com', 'bar.example.com'],
+  plugin  => 'nginx',
+}
+```
```
ekohl, line 133: Should this be an `@example Nginx authenticator` tag in `certonly.pp`? That way it ends up in `REFERENCE.md`. Not a blocker for this, but perhaps this should be done for all examples above

spec/classes/plugin/nginx_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      context 'with default parameters' do
+        it { is_expected.to compile.with_all_deps }
+
+        package_name =
+          if facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] == '7'
+            'python2-certbot-nginx'
+          else
+            'python3-certbot-nginx'
+          end
```
ekohl, line 24: I think `let` is cleaner for these multi line statements.
```suggestion
        let(:package_name) do
          if facts[:os]['family'] == 'RedHat' && facts[:os]['release']['major'] == '7'
            'python2-certbot-nginx'
          else
            'python3-certbot-nginx'
          end
        end
```

spec/classes/plugin/nginx_spec.rb
```diff
@@ -0,0 +1,38 @@
+require 'spec_helper'
+
+describe 'letsencrypt::plugin::nginx' do
+  on_supported_os.each do |os, facts|
```
ekohl, line 4: It's preferred to use `os_facts` to not shadow `facts`
```suggestion
  on_supported_os.each do |os, os_facts|
```

spec/classes/plugin/nginx_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    context "on #{os} based operating systems" do
```
ekohl, line 5: We generally keep it short
```suggestion
    context "on #{os}" do
```

README.md
nod0n: They are in the README.md right now, but I tried to add them in the `certonly.pp` to. Please have a look at https://github.com/voxpupuli/puppet-letsencrypt/pull/267/commits/576948a1ff73f70af8eb7e38bd06241fe985ffea

spec/classes/plugin/nginx_spec.rb
nod0n: I have to say, I don't know much about rspec and why let is the cleaner way^^ If you don't mind pointing me in the right direction, so I can learn something. :)

ekohl: https://stackoverflow.com/questions/5359558/when-to-use-rspec-let/5359979#5359979 does a good job and actually what I myself used. https://relishapp.com/rspec/rspec-core/docs/helper-methods/let-and-let is the official documentation.

In this particular case I'd prefer it because it lets you easily write multi-line statements.

manifests/certonly.pp
```diff
@@ -2,6 +2,46 @@
 #
 # This type can be used to request a certificate using the `certonly` installer.
 #
+# @example standalone authenticator
+#   letsencrypt::certonly { 'foo.example.com': }
+#
+# @example Apache authenticator
+#   letsencrypt::certonly { 'foo':
+#     domains => ['foo.example.com', 'bar.example.com'],
```
ekohl, line 10: While we're at it, I'd also consider splitting it:
```suggestion
# @example Multiple domains
#   letsencrypt::certonly { 'foo':
#     domains => ['foo.example.com', 'bar.example.com'],
#
# @example Apache authenticator
#   letsencrypt::certonly { 'foo.example.com':
```

This could be applied to all examples to keep them shorter. Again, a bit offtopic so feel free to ignore it for this PR.

manifests/plugin/nginx.pp
```diff
@@ -0,0 +1,14 @@
+# @summary install and configure the Let's Encrypt nginx plugin
+#
+# @param manage_package Manage the plugin package.
+# @param package_name The name of the package to install when $manage_package is true.
+class letsencrypt::plugin::nginx (
+  Boolean $manage_package = true,
+  String $package_name = 'python3-certbot-nginx',
```
ekohl, line 7: I'm always fan of non-empty string enforcement:
```suggestion
  String[1] $package_name = 'python3-certbot-nginx',
```

manifests/plugin/nginx.pp
```diff
@@ -0,0 +1,14 @@
...
+) {
+  if $manage_package {
+    package { $package_name:
+      ensure          => installed,
```
ekohl, line 11:
```suggestion
      ensure => installed,
```

spec/classes/plugin/nginx_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+          email => 'foo@example.com',
+        }
+        PUPPET
+      end
+
+      context 'with default parameters' do
+        it { is_expected.to compile.with_all_deps }
+
+        let(:package_name) do
+          if os_facts[:os]['family'] == 'RedHat' && os_facts[:os]['release']['major'] == '7'
```
ekohl: It is valid to refer to `facts` here. Then you use `let(:facts) { ...}`. So it will also take any overrides into account. That can be another reason to prefer `let`.

ekohl (commented):
Apologies for the scope creep. I hope you appreciate the learning part of it.

README.md
nod0n: Shall I remove some of them from the README.md or keep them in both places? It feels somewhat redundant to me, to have all examples in both files.

ekohl: In most repos we've removed them. You can keep a general section that says where they can be found in the reference for better visibility.

manifests/certonly.pp
nod0n: https://github.com/voxpupuli/puppet-letsencrypt/pull/267/commits/907aae5353c235b2eaeafd65ffa237d8ed64d56d

README.md
nod0n: In the README.md they have some comments (explanations). How do you add comments to examples in Puppet Strings?
In the past I think I just did:
```
# @example asdf
#   # explanation
#   code
#   ...
```

spec/classes/plugin/nginx_spec.rb
nod0n: Please excuse the probably stupid question, but why use `let(:facts) { ... }` in the first place. What is `let(:...) { ... }` good for? I can't read the complete rspec docs in 15 minutes, but maybe a chapter ;).

README.md
ekohl: I also used Puppet code comments in the same way. I'm not aware of any other method.

spec/classes/plugin/nginx_spec.rb
nod0n: Thank you! I think I understood. :)

nod0n: Now this makes some sense:
```puppet
describe 'with manage_package => false' do
          let(:params) { super().merge(manage_package: false, package_name: 'nginx-package') }

          it { is_expected.not_to contain_package('nginx-package') }
        end
```

nod0n: Should be improved in one of the commits.

README.md
nod0n: How do you like it now?
For some reason, I don't understand for what reason, the pipeline fails... Do you have an idea?

ekohl (approved):
I restarted the tests since the failures looked intermittent.

PR #271, opened 2022-01-05
create certificates from class parameter (hiera)

nod0n:
Support passing a hash of certificates from Hiera to the `letsencrypt` class.
This PR just iterates over the hash and creates `letsencrypt::certonly` for each entry.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -28,6 +28,7 @@
 # @param unsafe_registration A flag to allow using the 'register-unsafely-without-email' flag.
 # @param config_dir The path to the configuration directory.
 # @param key_size Size for the RSA public key
+# @param certificates A hash containing certificates. Each key is the title and each value is a hash, both passed to letsencrypt::certonly.
 # @param renew_pre_hook_commands Array of commands to run in a shell before obtaining/renewing any certificates.
 # @param renew_post_hook_commands Array of commands to run in a shell after attempting to obtain/renew certificates.
 # @param renew_deploy_hook_commands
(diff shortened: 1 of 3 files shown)


manifests/init.pp
```diff
@@ -52,30 +53,31 @@
...
+  String $config_file                = "${config_dir}/cli.ini",
+  Hash $config                       = { 'server' => 'https://acme-v02.api.letsencrypt.org/directory' },
+  String $cron_scripts_path          = "${facts['puppet_vardir']}/letsencrypt",
+  String $cron_owner_group           = 'root',
+  Boolean $manage_config             = true,
+  Boolean $manage_install            = true,
+  Boolean $agree_tos                 = true,
+  Boolean $unsafe_registration       = false,
+  Integer[2048] $key_size            = 4096,
+  Hash[String[1],Hash] $certificates = {},
```
traylenator, line 71: Can you add some spec tests for when this parameter is set that it does contain some `letsencrypt::certonly` entries.

nod0n: I'm very new to rspec but: https://github.com/voxpupuli/puppet-letsencrypt/pull/271/commits/9723ce55ffb1410b79b0f0ae4c7776fb53b14ece

spec/classes/letsencrypt_spec.rb
```diff
@@ -141,6 +141,13 @@
           it { is_expected.to contain_file('/foo/bar/baz').with(ensure: 'directory') }
         end
 
+        describe 'with list of certificates' do
+          let(:additional_params) { { certificates: { 'foo' => {}, 'a' => {} } } }
```
traylenator, line 145: Pass in some random parameters to `foo` for instance just to check it does what is expected.

nod0n:
rubocop is happy now. :)
should I squash the commits? Any other suggestions?

PR #275, opened 2022-01-20
move scripts into a separate class

nod0n:
letsencrypt::certonly required letsencrypt and when passing a
certificate list to $letsencrypt::certificates, it created a dependency
cycle.

This problem was discussed in slack: https://puppetcommunity.slack.com/archives/C0W1Y5VL0/p1642680497025900

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -252,7 +252,7 @@ define letsencrypt::certonly (
     environment => $environment,
     provider    => 'shell',
     require     => [
-      Class['letsencrypt'],
+      Exec['initialize letsencrypt'],
       File['/usr/local/sbin/letsencrypt-domain-validation'],
     ],
   }
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -141,6 +141,8 @@ define letsencrypt::certonly (
     fail("The 'webroot_paths' parameter must be specified when using the 'webroot' plugin")
   }
 
+  include letsencrypt::scripts
+
   # Wildcard-less title for use in file paths
   $title_nowc = regsubst($title, '^\*\.', '')
 
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,7 +93,6 @@ class letsencrypt (
   }
 
   contain letsencrypt::renew
-  include letsencrypt::scripts
 
   # TODO: do we need this command when installing from package?
   exec { 'initialize letsencrypt':
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -93,6 +93,7 @@ class letsencrypt (
   }
 
   contain letsencrypt::renew
+  include letsencrypt::scripts
 
   # TODO: do we need this command when installing from package?
   exec { 'initialize letsencrypt':
@@ -100,15 +101,6 @@ class letsencrypt (
     path        => $facts['path'],
     environment => $environment,
     refreshonly => true,
-  }
-
-  # Used in letsencrypt::certonly Exec["letsencrypt certonly ${title}"]
-  file { '/usr/local/sbin/letsencrypt-domain-validation':
-    ensure => file,
-    owner  => 'root',
-    group  => 'root',
-    mode   => '0500',
-    source => "puppet:///modules/${module_name}/domain-validation.sh",
   }
 
   $certificates.each |$certificate, $properties| {
--- a/manifests/scripts.pp
+++ b/manifests/scripts.pp
@@ -1,0 +1,16 @@
+# @summary Deploy helper scripts scripts
+#
+# @api private
+#
+class letsencrypt::scripts () {
+  assert_private()
+
+  # required in letsencrypt::certonly
+  file { '/usr/local/sbin/letsencrypt-domain-validation':
+    ensure  => file,
+    owner   => 'root',
+    group   => 'root',
+    mode    => '0500',
+    content => file("${module_name}/domain-validation.sh"),
+  }
+}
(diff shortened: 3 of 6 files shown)


spec/classes/letsencrypt_scripts_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'letsencrypt::scripts' do
+  on_supported_os.each do |os, facts|
+    let :facts do
+      facts
+    end
```
bastelfreak, line 9:
```suggestion
  on_supported_os.each do |os, os_facts|
    let :facts do
      os_facts
    end
```

that makes it easier to not mix up facts from on_supported_os.each and facts from the let block

manifests/scripts.pp
```diff
@@ -0,0 +1,16 @@
...
+class letsencrypt::scripts () {
+  assert_private()
+
+  # required in letsencrypt::certonly
+  file { '/usr/local/sbin/letsencrypt-domain-validation':
+    ensure => file,
+    owner  => 'root',
+    group  => 'root',
+    mode   => '0500',
+    source => "puppet:///modules/${module_name}/domain-validation.sh",
```
bastelfreak, line 14: while you update this, please use content to inline the file into the catalog. that reduces the number of http calls from puppet agent to puppetserver.
```suggestion
    content => file("${module_name}/domain-validation.sh"),
```

manifests/certonly.pp
```diff
@@ -252,7 +252,6 @@
     environment => $environment,
     provider    => 'shell',
     require     => [
-      Class['letsencrypt'],
```
ekohl, line 255: Isn't this still needed for certbot to installed and initialized?

spec/classes/letsencrypt_scripts_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'letsencrypt::scripts' do
+  on_supported_os.each do |os, os_facts|
+    let :facts do
+      os_facts
+    end
```
ekohl, line 9: This must be within `context "on #{os}"` or you get very weird results. You essentially only use the last set of facts for all OSes.

spec/classes/letsencrypt_scripts_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      context 'with default values' do
+        it { is_expected.to compile }
+
+        it 'contains the domain validation script' do
+          is_expected.to contain_file('/usr/local/sbin/letsencrypt-domain-validation').
+            with_ensure('file').
+            with_owner('root').
+            with_group('root').
+            with_mode('0500').
+            with_source('puppet:///modules/letsencrypt/domain-validation.sh')
```
ekohl, line 21: This will no longer pass. It should be `.with_content()` now that you use `content =>` (which I do like btw)

ekohl (changes requested):

spec/classes/letsencrypt_scripts_spec.rb
nod0n: Moved tests back into `letsencrypt_spec.rb`.

spec/classes/letsencrypt_scripts_spec.rb
nod0n: What regex/string should I check for?

spec/classes/letsencrypt_spec.rb
```diff
@@ -19,13 +19,13 @@
...
 
-          it 'contains File[/usr/local/sbin/letsencrypt-domain-validation]' do
+          it 'contains the domain validation script' do
             is_expected.to contain_file('/usr/local/sbin/letsencrypt-domain-validation').
               with_ensure('file').
               with_owner('root').
               with_group('root').
               with_mode('0500').
-              with_source('puppet:///modules/letsencrypt/domain-validation.sh')
+              with_content(%r{result=}) # TODO: what should be check?
```
ekohl, line 28: Probably sufficient to check for the shebang.

manifests/init.pp
```diff
@@ -93,6 +93,7 @@
   }
 
   contain letsencrypt::renew
+  include letsencrypt::scripts
```
ekohl, line 96: If this is only used in `certonly`, should that have the include instead?

manifests/certonly.pp
ekohl: This exec is the correct one for now. It would also be great if there was a test that validates all dependencies are correct. Dependency cycles are easy to detect in CI.

It looks like this code here checks the parameter:
https://github.com/voxpupuli/puppet-letsencrypt/blob/0eaecd43b1f7c588b5c1373a694bb4ce6fde801c/spec/classes/letsencrypt_spec.rb#L146-L156
Please add `it { is_expected.to compile.with_all_deps }` to it. I'd expect that to fail in master and your commit to solve it. Then we have a nice regression test.

spec/classes/letsencrypt_spec.rb
nod0n: done: https://github.com/voxpupuli/puppet-letsencrypt/pull/275/commits/cacb110a18d8f8fcd52f2897188f87caabef2c4c

manifests/certonly.pp
nod0n: done: https://github.com/voxpupuli/puppet-letsencrypt/pull/275/commits/e81f72210225ac3bf6b3f00cc436c52373c05065

manifests/init.pp
nod0n: done: 84f9150636e297c2447ed8c6f81d475c6917c2ae

ekohl (approved):
Looks good to me.

Just writing down my thoughts: I'm not entirely sure the separate class is really needed, just changing the `Class['letsencrypt']` to `Exec['initialize letsencrypt']` would probably be sufficient. However, I'm OK with the change since it does make things better. If you don't use `certonly`, you don't need the script. So :+1: to this.

bastelfreak (approved):

PR #277, opened 2022-04-11
Support logging cron output

treydock:

Allow logging letsencrypt renewal output.  So far hoping this will be useful when post_commands don't fire, so debug what's going wrong there.

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -22,7 +22,11 @@
 # @param manage_cron
 #   Indicating whether or not to schedule cron job for renewal.
 #   Runs daily but only renews if near expiration, e.g. within 10 days.
-# @param suppress_cron_output Redirect cron output to devnull
+# @param cron_output
+#   How to treat cron output
+#   `suppress` - Suppress all output
+#   `log` - Forward cron output to syslog
+#   undef - Do nothing with cron output (default)
 # @param cron_before_command Representation of a command that should be run before renewal command
 # @param cron_success_command Representation of a command that should be run if the renewal command succeeds.
 # @param cron_hour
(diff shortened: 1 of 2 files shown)


manifests/certonly.pp
```diff
@@ -58,6 +59,7 @@
   Array[String[1]]                          $environment          = [],
   Boolean                                   $manage_cron          = false,
   Boolean                                   $suppress_cron_output = false,
+  Boolean                                   $log_cron_output      = false,
```
smortex, line 62: These two are mutually exclusive.  Maybe add a test condition that raise an error if both or true?  Or maybe replace them with a single setting `Enum['suppress', 'log', 'ignore'] $cron_output = 'ignore'` ?

smortex (commented):
Please also consider adding / adjusting tests in spec/defines/letsencrypt_certonly_spec.rb near line 435 to ensure the new behavior does not break in the future.

manifests/certonly.pp
```diff
@@ -182,11 +186,18 @@
     $cron_script_ensure = $ensure ? { 'present' => 'file', default => 'absent' }
     $cron_ensure = $ensure
 
-    if $suppress_cron_output {
-      $croncommand = "${maincommand} > /dev/null 2>&1"
-    } else {
-      $croncommand = $maincommand
+    case $cron_output {
```
root-expert, line 189: I would use selector syntax here since it's more clean.

root-expert (approved):

manifests/certonly.pp
treydock: Made this change.

root-expert:
Please squash your commits before merging. Thanks!

treydock:
Commits squashed

smortex (approved):
LGTM!

PR #238, opened 2021-02-24
Add Cloudflare DNS plugin support

CallumBanbery:
#### Pull Request (PR) description

Adding feature support for the Certbot DNS Cloudflare plugin.

The plugin itself allows for two types of authentication, API token or Global API Key and corresponding Email so both are supported in this change.

Cloudflare recommends API tokens as they're more secure, so this module will prioritize token authentication over key authentication if both are provided.

#### This Pull Request (PR) fixes the following issues
n/a

PR #279, opened 2022-04-20
Add Cloudflare DNS plugin support

bkuebler:
Adding feature support for the Certbot DNS Cloudflare plugin.

The plugin itself allows for two types of authentication, API token or Global API Key and corresponding Email so both are supported in this change.

Cloudflare recommends API tokens as they're more secure, so this module will prioritize token authentication over key authentication if both are provided.

Replaces PR https://github.com/voxpupuli/puppet-letsencrypt/pull/238 or updates the missing tasks to fullfill
the merge failures.


data/os/Debian/10.yaml
```diff
@@ -1,3 +1,4 @@
 ---
 letsencrypt::plugin::dns_rfc2136::package_name: 'python3-certbot-dns-rfc2136'
 letsencrypt::plugin::dns_route53::package_name: 'python3-certbot-dns-route53'
+letsencrypt::plugin::dns_cloudflare::package_name: 'python3-certbot-dns-cloudflare'
```
ekohl, line 4: This inspired me to submit https://github.com/voxpupuli/puppet-letsencrypt/pull/280

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'spec_helper_acceptance'
+
+describe 'letsencrypt::plugin::dns_cloudflare' do
+  supported = case fact('os.family')
+              when 'Debian'
+                # Debian 9 has it in backports, Ubuntu started shipping in Bionic
```
ekohl, line 8: Debian 9 and Ubuntu 16.04 are no longer in metadata.json so they're no longer tested. I think this whole case can be simplified

bkuebler: Case reworked and obsolete content removed

data/os/Debian/10.yaml
bkuebler: @ekohl i added your idea https://github.com/voxpupuli/puppet-letsencrypt/pull/280 into this one here too.

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
```diff
@@ -3,15 +3,10 @@
...
-                # Debian 9 has it in backports, Ubuntu started shipping in Bionic
-                fact('os.release.major') != '9' && fact('os.release.major') != '16.04'
-              when 'RedHat'
-                true
-              else
-                false
-              end
+  supported = { 'Debian' => true,
+                'RedHat' => true,
+                'CentOS' => true }
```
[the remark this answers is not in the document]
bkuebler: Redundant definition removed.

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        config => {
+          'server' => 'https://acme-staging-v02.api.letsencrypt.org/directory',
+        },
+      }
+      class { 'letsencrypt::plugin::dns_cloudflare':
+        api_token => 'dummy-cloudflare-api-token',
+      }
+    PUPPET
+
+    if supported[fact('os.family')]
```
ekohl, line 23: I wonder if it's easier to use this:
```suggestion
    if ['Debian', 'RedHat'].include?(fact('os.family'))
```

Then you can drop the whole supported variable.

spec/classes/plugin/dns_cloudflare.rb
```diff
@@ -0,0 +1,59 @@
...
+          email => 'foo@example.com',
+        }
+        PUPPET
+      end
+      let(:package_name) do
+        osname = facts[:os]['name']
+        osrelease = facts[:os]['release']['major']
+        osfull = "#{osname}-#{osrelease}"
+        case osfull
+        when 'Debian-10', 'Ubuntu-18.04', 'Fedora-30', 'Fedora-31'
```
ekohl, line 22: This is odd. I'd expect this to fail on Debian 11, but the tests are green. This made me realize the tests are not executed at all. The filename must end with `_spec.rb`.

spec/classes/plugin/dns_cloudflare.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+require 'spec_helper'
+
+describe 'letsencrypt::plugin::dns_cloudflare' do
+  on_supported_os.each do |os, facts|
+    context "on #{os} based operating systems" do
+      let(:facts) { facts }
```
ekohl, line 8: It is now recommended to have a different name for the `let` block and the regular Ruby variable:
```suggestion
  on_supported_os.each do |os, os_facts|
    context "on #{os} based operating systems" do
      let(:facts) { os_facts }
```

spec/classes/plugin/dns_cloudflare.rb
bkuebler: Damn... didn't realized it, i just rebased the source branch form @CallumBanbery. Thanks i will have a look to fix this.

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
bkuebler: Will do 👍

spec/classes/plugin/dns_cloudflare.rb
bkuebler: 👍

data/os/Debian/10.yaml
ekohl: #280 is now merged, resulting in a merge conflict. Please rebase.

bkuebler: Done

spec/classes/plugin/dns_cloudflare_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+        }
+        PUPPET
+      end
+      let(:package_name) do
+        osname = facts[:os]['name']
+        osrelease = facts[:os]['release']['major']
+        osfull = "#{osname}-#{osrelease}"
+        if %w[RedHat-7 CentOS-7].include?(osfull)
+          'python2-certbot-dns-cloudflare'
+        elsif %w[Fedora].include?(osname) || %w[Debian RedHat].include?(facts[:os]['family'])
```
ekohl, line 23: Fedora is in the RedHat OS family
```suggestion
        elsif %w[Debian RedHat].include?(facts[:os]['family'])
```

data/os/Fedora.yaml
```diff
@@ -2,3 +2,4 @@
 letsencrypt::configure_epel: false
 letsencrypt::plugin::dns_rfc2136::package_name: 'python3-certbot-dns-rfc2136'
 letsencrypt::plugin::dns_route53::package_name: 'python3-certbot-dns-route53'
+letsencrypt::plugin::dns_cloudflare::package_name: 'python3-certbot-dns-cloudflare'
```
ekohl, line 5: This is partially pre-existing, but I think RedHat-family.yaml should cover all of this.

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    PUPPET
+
+    it 'installs letsencrypt and dns cloudflare plugin without error' do
+      apply_manifest(pp, catch_failures: true)
+    end
+
+    it 'installs letsencrypt and dns cloudflare idempotently' do
+      apply_manifest(pp, catch_changes: true)
+    end
+  end
```
ekohl, line 26: We have a shared example to make this shorter. See https://github.com/voxpupuli/voxpupuli-acceptance#an-idempotent-resource if you want to use this.

ekohl (approved):
Overall :+1: but some additional suggestions inline.

data/os/Fedora.yaml
bkuebler: Will remove it 👍

spec/classes/plugin/dns_cloudflare_spec.rb
bkuebler: Removed

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
bkuebler: Done, any hint about the failed tests on CentOS7? I can't find any issues why this happens.

root-expert: Seems it can't find `python2-certbot-dns-cloudflare` but it exists under EPEL.
Probably this is an ordering issue.

I see: `Notice: /Stage[main]/Epel/Epel::Rpm_gpg_key[EPEL-7]/Exec[import-EPEL-7]/returns: executed successfully`

after trying to install `python2-certbot-dns-cloudflare`

ekohl: I think for EL7 you somehow need to make sure EPEL is pulled in. Looks like the other DNS plugins have `require letsencrypt` which takes care of it:
https://github.com/voxpupuli/puppet-letsencrypt/blob/291f4b218230cf24477b74122c07703d6d119ce5/manifests/plugin/dns_rfc2136.pp#L27

bkuebler: Okay maybe i will add a Class dependency inside the acceptance test would be better:
```
Class['letsencrypt'] -> Class['letsencrypt::plugin::dns_cloudflare']
```
Because if i add `require letsencrypt` i will run into a dependency loop if i use the defined resource `letsencrypt::certonly`... maybe on the other plugins this will fail too and i assume no-one use this plugins currently..

ekohl: Perhaps it's then sufficient to `require letsencrypt::install`?

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+      class { 'letsencrypt' :
+        email  => 'letsregister@example.com',
+        config => {
+          'server' => 'https://acme-staging-v02.api.letsencrypt.org/directory',
+        },
+      }
+      class { 'letsencrypt::plugin::dns_cloudflare':
+        api_token => 'dummy-cloudflare-api-token',
+      }
+      Class['letsencrypt'] -> Class['letsencrypt::plugin::dns_cloudflare']
```
ekohl, line 18: I would feel better if this would be part of the class itself, not in the acceptance test. Otherwise it needs to be repeated for every use of it. Remember that `require $class` is the same as `include $class ; Class[$class] -> Class[$self]`.

bkuebler: Yes, but the class `letsencrypt::plugin::dns_cloudflare` should never be included if you use the `letsencrypt` plugin... it is enough to just use

```
include letsencrypt

letsencrypt::certonly { 'example.org':
  plugin => 'dns-cloudflare',
}
```

because the `letsencrypt::plugin::dns_cloudflare` class will be included then automatically. So a call like in the acceptance test is not required to use this plugin. If i add `require letsencrypt` into `letsencrypt::plugin::dns:_cloudflare` it will end into a dependency cycle at least if i use this module with puppet bolt. This was why i removed the require inside the plugin itself.

ekohl: As suggested in https://github.com/voxpupuli/puppet-letsencrypt/pull/279#discussion_r855025446, `require letsencrypt::install` might be sufficient then.

> If i add `require letsencrypt` into `letsencrypt::plugin::dns:_cloudflare` it will end into a dependency cycle at least if i use this module with puppet bolt. This was why i removed the require inside the plugin itself.

The spec test you added should also cover this.

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
bkuebler: Done... tried and works for me as well...

spec/acceptance/letsencrypt_plugin_dns_cloudflare_spec.rb
bkuebler: Okay ... changed it. Worked also with bolt in one of my projects. 👍

ekohl (approved):
Thanks!

PR #287, opened 2022-05-03
Fix registration after an unsafe one

### LINKED ISSUE #286: `register-unsafely-without-email` config key is kept when turning `unsafe_registration` back to `false`
## Affected Puppet, Ruby, OS and module versions/distributions

- Puppet: 6/7
- Distribution: Tested on Debian, all affected
- Module version: 8.0.2

## How to reproduce (e.g Puppet code you use)

First, apply a simple catalog using `unsafe_registration` sets to `true`
```puppet
class { 'letsencrypt':
  unsafe_registration => true,
}
```

Next apply a catalog using `unsafe_registration` sets to `false`
```puppet
class { 'letsencrypt':
  email => 'hello@example.com',
}
```

## What are you seeing

`register-unsafely-without-email` is set to `true` in config file.

## What behaviour did you expect instead

No `register-unsafely-without-email` in config file, or set to `false`.

## Any additional information you'd like to impart

PR is coming…

neomilium:
This PR allow user to register safely after using `unsafe_registration` flag.

Issue details at #286 

Fixes #286

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -29,13 +29,8 @@ class letsencrypt::config (
     $_config = $config
   }
 
-  $ensure_unsafe_registration = $unsafe_registration ? {
-    true  => present,
-    false => absent,
-  }
-
   ini_setting { "${config_file} register-unsafely-without-email true":
-    ensure  => $ensure_unsafe_registration,
+    ensure  => bool2str($unsafe_registration, 'present', 'absent'),
     path    => $config_file,
     section => '',
     setting => 'register-unsafely-without-email',
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -29,17 +29,23 @@ class letsencrypt::config (
     $_config = $config
   }
 
+  $ensure_unsafe_registration = $unsafe_registration ? {
+    true  => present,
+    false => absent,
+  }
+
+  ini_setting { "${config_file} register-unsafely-without-email true":
+    ensure  => $ensure_unsafe_registration,
+    path    => $config_file,
+    section => '',
+    setting => 'register-unsafely-without-email',
+    value   => true,
+    require => File[$config_dir],
+  }
+
   unless 'email' in $_config {
     if $unsafe_registration {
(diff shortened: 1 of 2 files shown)


ekohl:
Tests look red :(

neomilium:
Fixed, thanks @ekohl

manifests/config.pp
```diff
@@ -29,17 +29,23 @@
     $_config = $config
   }
 
+  $ensure_unsafe_registration = $unsafe_registration ? {
+    true  => present,
+    false => absent,
+  }
+
+  ini_setting { "${config_file} register-unsafely-without-email true":
+    ensure  => $ensure_unsafe_registration,
```
ekohl, line 38: You can consider `bool2str()`:
```suggestion
  ini_setting { "${config_file} register-unsafely-without-email true":
    ensure  => bool2str($unsafe_registration, 'present', 'absent'),
```

manifests/config.pp
```diff
@@ -29,17 +29,23 @@
...
+    path    => $config_file,
+    section => '',
+    setting => 'register-unsafely-without-email',
+    value   => true,
+    require => File[$config_dir],
```
ekohl, line 43: It feels to me that this is something that ini_setting ought to do, so I checked and it's there:
https://github.com/puppetlabs/puppetlabs-inifile/blob/a0ef2bdc68dcb01713f1f3e50a44824e053927a3/lib/puppet/type/ini_setting.rb#L155-L157

It was added in https://github.com/puppetlabs/puppetlabs-inifile/commit/a2b46d20f298b2f3a861a542c2894031766e2829 which is version 2.5.0 and higher. If you bump that in `metadata.json` you can leave this line out.

On the other hand, you are just moving things around.

ekohl (approved):
Feel free to either address the comments or merge without.

manifests/config.pp
neomilium: Thanks

ekohl (approved):

Issue #250, opened 2021-09-28
Recent update causes problems with the nginx plugin


hboetes:
On this host, letsencrypt did work fine with a workaround I had in place, which I lost, so I have to start digging again:
```
2021-09-28 05:32:46,599:DEBUG:certbot.main:certbot version: 0.31.0
2021-09-28 05:32:46,600:DEBUG:certbot.main:Arguments: ['-q']
2021-09-28 05:32:46,602:DEBUG:certbot.main:Discovered plugins: PluginsRegistry(PluginEntryPoint#manual,PluginEntryPoint#nginx,PluginEntryPoint#null,PluginEntryPoint#standalone,PluginEntryPoint#webroot)
2021-09-28 05:32:46,611:DEBUG:certbot.log:Root logging level set at 30
2021-09-28 05:32:46,611:INFO:certbot.log:Saving debug log to /var/log/letsencrypt/letsencrypt.log
2021-09-28 05:32:46,626:DEBUG:certbot.plugins.selection:Requested authenticator <certbot.cli._Default object at 0x7fdd938cdd30> and installer <certbot.cli._Default object at 0x7fdd938cdd30>
2021-09-28 05:32:46,626:DEBUG:certbot.cli:Var server=https://acme-v02.api.letsencrypt.org/directory (set by user).
2021-09-28 05:32:46,626:DEBUG:certbot.cli:Var server=https://acme-v02.api.letsencrypt.org/directory (set by user).
2021-09-28 05:32:46,626:DEBUG:certbot.cli:Var account={'server'} (set by user).
2021-09-28 05:32:46,642:INFO:certbot.renewal:Cert not yet due for renewal
2021-09-28 05:32:46,652:DEBUG:certbot.plugins.selection:Requested authenticator nginx and installer None
2021-09-28 05:32:46,652:DEBUG:certbot.renewal:no renewal failures
2021-09-28 12:41:02,591:DEBUG:certbot.main:certbot version: 0.31.0
2021-09-28 12:41:02,592:DEBUG:certbot.main:Arguments: ['--keep-until-expiring', '--text', '--agree-tos', '--non-interactive', '--rsa-key-size', '4096', '-a', 'nginx', '--cert-name', 'quigon.example.com', '-d', 'quigon.example.com']
[...]

PR #288, opened 2022-05-11
Add environment parameter to renew cron

gmenuel:
This PR add an option for adding environment to the global renew cron, this is useful for example with the nginx certbot plugin which somehow can't find nginx if the PATH in the cron doesn't include /usr/sbin.

This PR is similar to #189 which doesn't seem really active. 

Fixes: #250 (well kind of, it doesn't directly fix this, but it allows to add the PATH in the cron which will fix the problem).

### DIFF
--- a/manifests/renew.pp
+++ b/manifests/renew.pp
@@ -26,6 +26,9 @@
 # @param cron_monthday
 #   Optional string, integer or array of monthday(s) the renewal command should
 #   run. E.g. '2-30/2' to run on even days. Default: Every day.
+# @param cron_environment
+#   Optional string or array of environment variables the renewal command should have.
+#   E.g. PATH=/sbin:/usr/sbin:/bin:/usr/bin
 #
 class letsencrypt::renew (
   Variant[String[1], Array[String[1]]] $pre_hook_commands    = $letsencrypt::renew_pre_hook_commands,
@@ -36,6 +39,7 @@ class letsencrypt::renew (
   Letsencrypt::Cron::Hour              $cron_hour            = $letsencrypt::renew_cron_hour,
   Letsencrypt::Cron::Minute            $cron_minute          = $letsencrypt::renew_cron_minute,
   Letsencrypt::Cron::Monthday          $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  Variant[String[1], Array[String[1]]] $cron_environment     = $letsencrypt::renew_cron_environment,
 ) {
   # Directory used for Puppet-managed renewal hooks. Make sure old unmanaged
   # hooks in this directory are purged. Leave custom hooks in the default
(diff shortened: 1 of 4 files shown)


hboetes:
Thanks!

manifests/renew.pp
```diff
@@ -26,16 +26,20 @@
...
-  Letsencrypt::Cron::Monthday          $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  Variant[String[1], Array[String[1]]]  $pre_hook_commands    = $letsencrypt::renew_pre_hook_commands,
+  Variant[String[1], Array[String[1]]]  $post_hook_commands   = $letsencrypt::renew_post_hook_commands,
+  Variant[String[1], Array[String[1]]]  $deploy_hook_commands = $letsencrypt::renew_deploy_hook_commands,
+  Array[String[1]]                      $additional_args      = $letsencrypt::renew_additional_args,
+  Enum['present', 'absent']             $cron_ensure          = $letsencrypt::renew_cron_ensure,
+  Letsencrypt::Cron::Hour               $cron_hour            = $letsencrypt::renew_cron_hour,
+  Letsencrypt::Cron::Minute             $cron_minute          = $letsencrypt::renew_cron_minute,
+  Letsencrypt::Cron::Monthday           $cron_monthday        = $letsencrypt::renew_cron_monthday,
+  Variant[String, Array[String], Undef] $cron_environment     = $letsencrypt::renew_cron_environment,
```
ekohl, line 42: Would it make sense for the type to default to an empty array and disallow `Undef`? I'm not sure if that makes a difference for the cron type.

ekohl (approved):

manifests/renew.pp
gmenuel: I've tested with [] and it removed the environment if it was previously setted. With the option unset (or undef) it doesn't remove the environment. But I think that removing the environment might be the right thing to do.

DJClean:
Just a quick question, is there anything holding this PR back?
Running into issues where I need to set the PATH for this cron to add some dirs. Which this PR would solve
If there are issues I don't mind taking a look and trying to fix it.

manifests/renew.pp
```diff
@@ -26,6 +26,9 @@
 # @param cron_monthday
 #   Optional string, integer or array of monthday(s) the renewal command should
 #   run. E.g. '2-30/2' to run on even days. Default: Every day.
+# @param cron_environment
+#   Optional string or array of environments(s) the renewal command should have.
```
kenyon, line 30:
```suggestion
#   Optional string or array of environment variables the renewal command should have.
```

gmenuel:
Merge conflicts resolved.

kenyon (approved):

manifests/renew.pp
kenyon: > I've tested with [] and it removed the environment if it was previously setted. With the option unset (or undef) it doesn't remove the environment. But I think that removing the environment might be the right thing to do.

@gmenuel I wonder then if we should mark this as a breaking change, since it would remove manual changes to the cron job.

gmenuel: @kenyon yes you are right, it is probably better to tag this as a breaking change.

PR #289, opened 2022-05-11
add option for removing cron job provided by distribution packages

### LINKED ISSUE #164: /etc/cron.d/certbot cron file should be absent when manage_cron is set to true on Debian
In my opinion the `/etc/cron.d/certbot` (default cron job for running certbot twice per day on Debian with official cerbot package) file should be absent when one sets `manage_cron` to `true` as both can/will interfere.

I did not check on other distributions but this might also be an issue with other distros.

For your reference here is the contact of that `/etc/cron.d/certbot` file on Debian 9:

```
# /etc/cron.d/certbot: crontab entries for the certbot package
#
# Upstream recommends attempting renewal twice a day
#
# Eventually, this will be an opportunity to validate certificates
# haven't been revoked, etc.  Renewal will only occur if expiration
# is within 30 days.
SHELL=/bin/sh
PATH=/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin

0 */12 * * * root test -x /usr/bin/certbot -a \! -d /run/systemd/system && perl -e 'sleep int(rand(3600))' && certbot -q renew
```

What do you think?

gmenuel:

This PR aims at removing the system cron or systemd timer when the renew cron is managed by puppet (this doesn't handle certonly cron). If we don't do that on certains distro (for exemple Debian), a cron will be automaticaly added by the package and might renew the certificate before the puppet cron, and thus the puppet renew hooks won't be executed.

Fixes #164

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -39,6 +39,7 @@
 #   - $RENEWED_DOMAINS: A space-delimited list of renewed certificate domains.
 #                       Example: "example.com www.example.com"
 # @param renew_additional_args Array of additional command line arguments to pass to 'certbot renew'.
+# @param renew_disable_distro_cron Boolean, set to true to disable the cron created by the distro package
 # @param renew_cron_ensure Intended state of the cron resource running certbot renew.
 # @param renew_cron_hour
 #   Optional string, integer or array of hour(s) the renewal command should run.
@@ -91,6 +92,7 @@ class letsencrypt (
   Letsencrypt::Cron::Minute $renew_cron_minute = fqdn_rand(60),
   Letsencrypt::Cron::Monthday $renew_cron_monthday = '*',
   Variant[String[1], Array[String[1]]] $renew_cron_environment = [],
+  Boolean                              $renew_disable_distro_cron = false,
   # define default hooks for all certonly defined resources
   Array[String[1]] $certonly_pre_hook_commands = [],
   Array[String[1]] $certonly_post_hook_commands = [],
--- a/manifests/renew.pp
+++ b/manifests/renew.pp
@@ -16,6 +16,9 @@
 #   - $RENEWED_DOMAINS: A space-delimited list of renewed certificate domains.
 #                       Example: "example.com www.example.com"
 # @param additional_args Array of additional command line arguments to pass to 'certbot renew'.
+# @param disable_distro_cron Boolean, set to true to disable the cron created by the distro package
+# @param distro_renew_cron_file Optional Unixpath, if set and if disable_distro_cron is true this file will be deleted (unless systemd is used)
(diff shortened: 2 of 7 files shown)


manifests/renew.pp
```diff
@@ -84,4 +91,23 @@
     minute   => $cron_minute,
     monthday => $cron_monthday,
   }
+
+  if ($disable_distro_cron) {
+    if ($facts['service_provider'] == 'systemd') {
+      if $distro_renew_timer {
```
treydock, line 97: I'd recommend changing this to be ANDed condition to avoid 3 level if condition.
```suggestion
  if $disable_distro_cron and $distro_renew_timer and $facts['service_provider'] == 'systemd' {
```

Would need to adjust depth of resources and remove extra closing brackets.

treydock:
@gmenuel Please address the merge conflicts.

spec/classes/letsencrypt_spec.rb
```diff
@@ -211,6 +211,41 @@
...
+            end
+          end
+
+          describe 'renew_cron_ensure and disable_distro_cron (without systemd)' do
+            let(:additional_params) do
+              { renew_cron_ensure: 'present' }
+            end
+
+            it do
+              case facts[:osfamily]
```
bastelfreak, line 240: can you please don't use the legacy facts?
```suggestion
              case facts[:os]['family']
```

gmenuel: Done

gmenuel:
I've updated the cron file name for freeBSD. I've don't have a freeBSD to test that it is correct (and I'm not particulary familiar with it) so if anyone can check that this is correct it would be great :)

gmenuel:
If anyone is still interested by this PR, I rebased it yesterday.

manifests/init.pp
```diff
@@ -91,6 +92,7 @@
   Letsencrypt::Cron::Minute $renew_cron_minute = fqdn_rand(60),
   Letsencrypt::Cron::Monthday $renew_cron_monthday = '*',
   Variant[String[1], Array[String[1]]] $renew_cron_environment = [],
+  Boolean                              $renew_disable_distro_cron = true,
```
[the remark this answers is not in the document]
TheMeier: or just use `false` as default. easy enough to golablly set it in hiera for a given control repo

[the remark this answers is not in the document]
gmenuel: I have changed it to false for the moment, otherwise yeah it would be backwards incompatible (even if I think that disabling the distro cron would be the intended behavior, since otherwise the post renew hooks might not get executed)

kenyon (approved):

PR #291, opened 2022-06-16
Fix misuse of fqdn_rand_string

smokris:
According to the puppetlabs-stdlib documentation,
fqdn_rand_string()'s second argument is the character set to use
when generating the random string, and its third argument is
the seed.  The certonly class incorrectly passes the domain name
as the second argument, causing fqdn_rand_string() to use the
host's fqdn as the seed, resulting in more-frequent-than-necessary
hash collisions.

(no issue)

### DIFF
--- a/manifests/certonly.pp
+++ b/manifests/certonly.pp
@@ -131,7 +131,7 @@ define letsencrypt::certonly (
   Optional[String[1]]                       $cron_success_command = undef,
   Array[Variant[Integer[0, 59], String[1]]] $cron_monthday        = ['*'],
   Variant[Integer[0,23], String, Array]     $cron_hour            = fqdn_rand(24, $title),
-  Variant[Integer[0,59], String, Array]     $cron_minute          = fqdn_rand(60, fqdn_rand_string(10, $title)),
+  Variant[Integer[0,59], String, Array]     $cron_minute          = fqdn_rand(60, $title),
   Stdlib::Unixpath                          $config_dir           = $letsencrypt::config_dir,
   Variant[String[1], Array[String[1]]]      $pre_hook_commands    = [],
   Variant[String[1], Array[String[1]]]      $post_hook_commands   = [],
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -283,7 +283,7 @@ Data type: `Any`
 Optional string, integer or array of minute(s) the renewal command should
 run. E.g. 0 or '00' or [0,30].
 
-Default value: `fqdn_rand(60, fqdn_rand_string(10))`
+Default value: `fqdn_rand(60)`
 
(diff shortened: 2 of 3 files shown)


kenyon (approved):

manifests/certonly.pp
```diff
@@ -131,7 +131,7 @@
   Optional[String[1]]                       $cron_success_command = undef,
   Array[Variant[Integer[0, 59], String[1]]] $cron_monthday        = ['*'],
   Variant[Integer[0,23], String, Array]     $cron_hour            = fqdn_rand(24, $title),
-  Variant[Integer[0,59], String, Array]     $cron_minute          = fqdn_rand(60, fqdn_rand_string(10, $title)),
+  Variant[Integer[0,59], String, Array]     $cron_minute          = fqdn_rand(60, fqdn_rand_string(10, undef, $title)),
```
smortex, line 134: Is the usage of `fqdn_rand_string()` needed?  I feel like we just want the `$cron_minute` to be different for each resource so we can do:

```suggestion
  Variant[Integer[0,59], String, Array]     $cron_minute          = fqdn_rand(60, $title),
```

See https://puppet.com/docs/puppet/7/function.html#fqdn-rand

smortex (changes requested):
I think you are right and the passed parameters where wrong. That being said, I think we do not need the `fqdn_rand_string()` at all?

manifests/certonly.pp
smokris: Hmm, good question.  The `fqdn_rand_string()` call was added in https://github.com/voxpupuli/puppet-letsencrypt/commit/aeed58ae, whose commit message says:

> fixed randomness if the domain is almost same, like example.com and www.example.com/

I performed some tests:

```console
$ puppet apply -e 'notice(fqdn_rand(24, "example.com")) notice(fqdn_rand(60, "example.com"))'
Notice: Scope(Class[main]): 21
Notice: Scope(Class[main]): 39

$ puppet apply -e 'notice(fqdn_rand(24, "www.example.com")) notice(fqdn_rand(60, "www.example.com"))'
Notice: Scope(Class[main]): 8
Notice: Scope(Class[main]): 30

$ puppet apply -e 'notice(fqdn_rand(24, "www.example.com/")) notice(fqdn_rand(60, "www.example.com/"))'
Notice: Scope(Class[main]): 10
Notice: Scope(Class[main]): 52
```

Each of the hour/minute values is unique.  At least with the current implementation of `fqdn_rand()`, it looks like that's no longer a concern, so I'll go ahead and remove the `fqdn_rand_string()`.

smokris: 28d67f5

smortex (approved):
LGTM!

smokris:
Thanks, @smortex!

PR #295, opened 2022-08-26
Add support for the Certbot Gandi plugin

cible:
Adding feature support for the certbot-plugin-gandi created by @obynio

The plugin use the Production API key.

N/A

### DIFF
--- a/data/os/Ubuntu/24.04.yaml
+++ b/data/os/Ubuntu/24.04.yaml
@@ -1,0 +1,2 @@
+---
+letsencrypt::plugin::dns_gandi::package_name: python3-certbot-dns-gandi


data/FreeBSD-family.yaml
```diff
@@ -5,3 +5,4 @@ letsencrypt::cron_owner_group: 'wheel'
 letsencrypt::plugin::dns_rfc2136::package_name: 'py38-certbot-dns-rfc2136'
 letsencrypt::plugin::dns_route53::package_name: 'py38-certbot-dns-route53'
 letsencrypt::plugin::dns_cloudflare::package_name: 'py38-certbot-dns-cloudflare'
+letsencrypt::plugin::dns_gandi::package_name: 'certbot-plugin-gandi'
```
smortex, line 8: There are currently no FreeBSD port for this :-/

The name would probably be `py38-certbot-dns-gandi` if there was, and in fact `py39-certbot-dns-gandi` because the default version of Python changed recently on FreeBSD.

smortex: I opened #296 to adjust the FreeBSD version.

cible: Do you think it's a good idea to install the package with pip? I'm not sure on how to install them on freebsd and centos. I will ask help to a colleague

smortex: As a FreeBSD user, I would prefer to use a package rather installing using pip.  Anybody can submit ports to FreeBSD, and they are ultimately available as packages.  My guess is that if the port does not exist, nobody needed it so much they did a port for it, and maybe it is not worth it?

I would personally not set a `package_name` for FreeBSD in Hiera and make `$letsencrypt::plugin::dns_gandi::package_name` mandatory so that it fails hard if someone attempt to use it on FreeBSD.

data/os/Debian/11.yaml
```diff
@@ -0,0 +1,3 @@
+---
+letsencrypt::plugin::dns_gandi::package_provider: apt
```
ekohl, line 2: Why do you need to set the package provider? That's rarely needed.

manifests/plugin/dns_gandi.pp
```diff
@@ -0,0 +1,55 @@
...
+      package { $package_name:
+        ensure   => installed,
+        provider => $package_provider,
+      }
+    }
+    else {
+      package { $package_name:
+        ensure => installed,
+      }
+    }
```
ekohl, line 35: You can just pass `undef` along:
```suggestion
    package { $package_name:
      ensure   => installed,
      provider => $package_provider,
    }
```

manifests/plugin/dns_gandi.pp
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  if $api_key {
+    $ini_vars = {
+      'certbot_plugin_gandi:dns_api_key' => $api_key,
+    }
+  } else {
+    fail('api_key not provided for certbot dns gandi plugin.')
+  }
```
ekohl, line 44: The data type guarantees it's set
```suggestion
  $ini_vars = {
    'certbot_plugin_gandi:dns_api_key' => $api_key,
  }
```

data/RedHat-family.yaml
```diff
@@ -3,3 +3,4 @@ letsencrypt::configure_epel: true
 letsencrypt::plugin::dns_rfc2136::package_name: 'python3-certbot-dns-rfc2136'
 letsencrypt::plugin::dns_route53::package_name: 'python3-certbot-dns-route53'
 letsencrypt::plugin::dns_cloudflare::package_name: 'python3-certbot-dns-cloudflare'
+letsencrypt::plugin::dns_gandi::package_name: 'certbot-plugin-gandi'
```
ekohl, line 6: This doesn't look correct.

treydock:
@cible Is this ready for review?  The pull request is currently marked as a draft.

manifests/plugin/dns_gandi.pp
```diff
@@ -0,0 +1,43 @@
...
+  Stdlib::Absolutepath $config_file      = "${letsencrypt::config_dir}/dns-gandi.ini",
+  Boolean              $manage_package   = true,
+) {
+  require letsencrypt
+
+  if $manage_package {
+    if ! $package_name {
+      fail('No package name provided for certbot dns gandi plugin.')
+    }
+
```
ekohl, line 24: Given `$package_name` is a `String[1]`, this code will never be hit
```suggestion
```

manifests/plugin/dns_gandi.pp
```diff
@@ -0,0 +1,43 @@
...
+#
+# @param api_key Gandi production api key secret. You can get it in you security tab of your account
+# @param package_name The name of the package to install when $manage_package is true.
+# @param config_file The path to the configuration file.
+# @param manage_package Manage the plugin package.
+#
+class letsencrypt::plugin::dns_gandi (
+  String[1]            $api_key,
+  String[1]  $package_name,
+  Optional[String[1]]  $package_provider = undef,
```
ekohl, line 14: This is an unused parameter. Shouldn't it be passed to the `package` resource?

manifests/plugin/dns_gandi.pp
```diff
@@ -0,0 +1,43 @@
...
+) {
+  require letsencrypt
+
+  if $manage_package {
+    if ! $package_name {
+      fail('No package name provided for certbot dns gandi plugin.')
+    }
+
+    package { $package_name:
+      ensure   => installed,
```
ekohl, line 26: Does it make sense to ensure the package installation happens before the ini file is written?
```suggestion
      ensure => installed,
      before => File[$config_file],
```

spec/spec_helper_acceptance.rb
```diff
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require 'voxpupuli/acceptance/spec_helper_acceptance'
+require 'spec_helper_local'
```
ekohl, line 4: Why would you introduce a separate file for this? The file isn't managed now so you might as well include it here.

data/os/Debian/11.yaml
cible: You're right, I will remove it

spec/classes/plugin/dns_gandi_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+            is_expected.to compile.with_all_deps
+          end
+        end
+
+        describe 'with manage_package => true' do
+          let(:params) { super().merge(manage_package: true) }
+
+          it do
+            if package_name.nil?
+              is_expected.not_to compile
```
ekohl, line 43: A better way is to test for the error:
```suggestion
              is_expected.to compile.and_raise_error(/A Matcher For The Error/)
```

spec/defines/letsencrypt_certonly_spec.rb
```diff
@@ -207,6 +207,45 @@ class { 'letsencrypt::plugin::dns_cloudflare':
...
+            email      => 'foo@example.com',
+            config_dir => '/etc/letsencrypt',
+          }
+          class { 'letsencrypt::plugin::dns_gandi':
+            package_name => 'irrelevant',
+          }
+          PUPPET
+        end
+
+        it { is_expected.not_to compile.with_all_deps }
```
ekohl, line 246: Here too it's better to compile and test for a specific error

cible: For this one, the module won't work with an empty api_key and I don't know how to handle the error: 
```
  13) letsencrypt::certonly on fedora-36-x86_64 based operating systems with dns-gandi plugin without api_key is expected to fail to compile and raise an error matching /\/expects a value for parameter 'api_key'\//
      Failure/Error: it { is_expected.to compile.and_raise_error(%r{/expects a value for parameter 'api_key'/}) }
        error during compilation: Evaluation Error: Error while evaluating a Resource Statement, Class[Letsencrypt::Plugin::Dns_gandi]: expects a value for parameter 'api_key' (line: 6, column: 11
```
`it { is_expected.to compile.and_raise_error(%r{/expects a value for parameter 'api_key'/}) }` is not correct?

ekohl: The `/` is redundant. In Ruby you can define a regex as `/my regex/` or `%r{my regex}` but `%r{/my regex/}` means you expect `/` in the output. This is useful if you want to test `/var/lib is empty` is in the error. Compare `%r{/var/lib is empty}` to `/\/var\/lib is empty/`.

cible:
It seem ok for me. Sorry for being so slow and bad with ruby. Did I have to do something else? rebase?

bastelfreak:
@cible yes, this needs a rebase. I think we can merge if afterwards :+1:

PR #301, opened 2022-11-18
Drop Fedora 32, add Fedora 36 support

treydock:
Fedora 32 was EOL a while ago, so switch to latest "stable" release of Fedora.

### DIFF
--- a/spec/classes/plugin/dns_rfc2136_spec.rb
+++ b/spec/classes/plugin/dns_rfc2136_spec.rb
@@ -19,7 +19,7 @@ describe 'letsencrypt::plugin::dns_rfc2136' do
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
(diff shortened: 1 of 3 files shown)


spec/classes/plugin/dns_rfc2136_spec.rb
```diff
@@ -19,7 +19,7 @@ class { 'letsencrypt':
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
-        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-22.04', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-32'
+        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-22.04', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-36'
```
bastelfreak, line 22: we should really turn this into the default block

bastelfreak (approved):

PR #306, opened 2022-12-20
use gid 0 rather than group 'root' to let the module work on BSDs.

rtprio:
On FreeBSD and OpenBSD gid 0  is 'wheel', not 'root' and it will fail.
Using gid 0 corrects this simply.

I didn't find an issue for this, but the PR #231 is similar.

### DIFF
--- a/.msync.yml
+++ b/.msync.yml
@@ -2,4 +2,4 @@
 # Managed by modulesync - DO NOT EDIT
 # https://voxpupuli.org/docs/updating-files-managed-with-modulesync/
 
-modulesync_config_version: '5.3.0'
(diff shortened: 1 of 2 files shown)


rtprio:
Ah, nuts. I forgot the tests; will update.

Is this change allowed or is it better design to set the group based on OS?

kenyon:
There is precedent for putting these names in hiera, e.g., https://github.com/voxpupuli/puppet-letsencrypt/blob/4c9394291416fa8fde9dea3c8ef4a80478befac1/data/FreeBSD-family.yaml#L4 so maybe keep doing that instead?

PR #309, opened 2023-02-12
Remove unnecessary exec

deric:
Executing `letsencrypt -h` is not needed when installing binaries from a package.

### DIFF
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -81,7 +81,6 @@ class letsencrypt (
 ) {
   if $manage_install {
     contain letsencrypt::install # lint:ignore:relative_classname_inclusion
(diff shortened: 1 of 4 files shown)


manifests/init.pp
```diff
@@ -81,27 +81,17 @@
...
   $command = $package_command
 
   if $manage_config {
     contain letsencrypt::config # lint:ignore:relative_classname_inclusion
-    Class['letsencrypt::config'] -> Exec['initialize letsencrypt']
   }
 
   contain letsencrypt::renew
 
-  # TODO: do we need this command when installing from package?
```
ekohl, line 97: Glad a TODO is solved :)

ekohl (approved):

PR #310, opened 2023-03-30
Avoid race condition when renewing certificates

deric:

This PR makes sure that only single certificate can be processed at the same time. Make sure to obtain exclusive lock before executing renewal command.

`flock` command is used to obtain exclusive lock (with 30 seconds timeout).

Resolve errors like this, when randomized cron job time is not enough:
```
Another instance of Certbot is already running.
```


templates/renew-script.sh.erb
```diff
@@ -1,5 +1,10 @@
-#!/bin/sh
+#!/bin/bash
```
smortex, line 1: `metadata.json` says the module work on some non-linux systems where you can't count on `bash(1)` (and if available is not in `/bin`). So this script should be kept portable.

```suggestion
#!/bin/sh
```

templates/renew-script.sh.erb
```diff
@@ -1,5 +1,10 @@
...
 <%- @environment.each do |environment| -%>
 export <%= environment %>
 <%- end -%>
+if [[ -f '/usr/share/certbot_lock.sh' ]]; then
```
smortex, line 5: `[[` is a bashism, use `[`:
```suggestion
if [ -f '/usr/share/certbot_lock.sh' ]; then
```

files/certbot_lock.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
```
smortex, line 1: Only add shebangs to programs.

files/certbot_lock.sh
```diff
@@ -0,0 +1,14 @@
...
+LOCKFILE="/var/lock/certbot"
+LOCKFD=99
+# private
+function _lock()          { flock "$@" $LOCKFD; }
```
smortex, line 5: `flock(1)` does not appear to be something standard. On FreeBSD for example, we have `lockf(1)` (which seems to serve the same purpose) but no `flock(1)`.

The whole logic seems over-engineered, `flock(1)` and `lockf(1)` both seems to support a simple syntax with a file and a command where the file is locked while the command run. Maybe you can try to find if the system has `lockf(1)` / `flock(1)` (e.g. using the `type` built-in command) and use it if found; and otherwise fallback to not locking?

smortex (changes requested):
Unfortunately it needs some work. I added in-line comments regarding the concerns I have reviewing the change.  Feel free to ask if you need more details! Thank you

templates/renew-script.sh.erb
deric: Fixed.

templates/renew-script.sh.erb
deric: Fixed.

files/certbot_lock.sh
deric: On Debian `flock` is part of `util-linux` which is [essential package](https://packages.debian.org/bullseye/util-linux) that is automatically installed. But of course other distributions might take very different approach.

I've tried to keep the code as simple as possible, but if you have better suggestion, feel free to improve it.

vchepkov: set -o noclobber not enough?

ekohl (commented):
I still question why this module has such a complex renewal process. Certbot itself has hooks to run after certain points. Why can't it rely on that and use `certbot renew`? That's already by default running in cron / systemd timer in many Linux packages and makes this module much less invasive.

Quoting https://eff-certbot.readthedocs.io/en/stable/using.html#renewing-certificates

> You can also specify hooks by placing files in subdirectories of Certbot’s configuration directory. Assuming your configuration directory is /etc/letsencrypt, any executable files found in /etc/letsencrypt/renewal-hooks/pre, /etc/letsencrypt/renewal-hooks/deploy, and /etc/letsencrypt/renewal-hooks/post will be run as pre, deploy, and post hooks respectively when any certificate is renewed with the renew subcommand. These hooks are run in alphabetical order and are not run for other subcommands. (The order the hooks are run is determined by the byte value of the characters in their filenames and is not dependent on your locale.)

https://eff-certbot.readthedocs.io/en/stable/using.html#id5 also says it already has locks in place.

Doesn't this move the whole module in the wrong direction by inventing more custom code?

files/certbot_lock.sh
deric: @vchepkov No, it's not enough. We don't want immediate error (that's already covered by certbot), we need lock & wait for releasing previous lock (if exists).

@smortex I'm not familiar with FreeBSD, but in FreeBSD 13.8 `flock` [seems to be available](https://man.freebsd.org/cgi/man.cgi?query=flock&sektion=1&manpath=FreeBSD+13.2-RELEASE+and+Ports). `lockf` is only a C function (same on Linux).

deric:
@ekohl Yes, there are locking mechanisms in `cerbot`, but the process will end with an error instead of waiting for releasing the lock.

I agree it's pretty complex. Part of the problem is that:

```puppet
$cron_minute = fqdn_rand(60, $title)
$cron_hour     = fqdn_rand(24, $title)
```
doesn't provide much flexibility. With ~20-30 certificates you can easily run into collisions. If you have only a few certificates you should be fine (or you could define those times manually).

ekohl:
@deric my whole suggestion was to use the command `certbot renew` instead of one cronjob per certificate. `certbot` can figure out what needs renewel on its own. It's a big refactor for the whole module, but it would greatly simplify things.

bastelfreak:
I agree with @ekohl . In the past I also had one systemd timer per certificate but that doesn't make much sense. certbot itself is really flexible with different hooks and can deal multiple certificates.

deric:
@ekohl You're right, reducing per-site cron jobs to a single cron job would be definitely a better option.

ekohl:
There's even an additional advantage: if multiple certificates are renewed at the same time you can reload/restart services which consume multiple certificates (Apache, nginx) only once instead of for each certificate.

PR #312, opened 2023-04-04
Add Rocky 8 support

bastelfreak:

### DIFF
--- a/spec/classes/plugin/dns_rfc2136_spec.rb
+++ b/spec/classes/plugin/dns_rfc2136_spec.rb
@@ -19,7 +19,7 @@ describe 'letsencrypt::plugin::dns_rfc2136' do
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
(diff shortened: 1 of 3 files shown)


spec/classes/plugin/dns_rfc2136_spec.rb
```diff
@@ -19,7 +19,7 @@ class { 'letsencrypt':
         osrelease = facts[:os]['release']['major']
         osfull = "#{osname}-#{osrelease}"
         case osfull
-        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-22.04', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-36'
+        when 'Debian-10', 'Debian-11', 'AlmaLinux-8', 'RedHat-8', 'Ubuntu-22.04', 'Ubuntu-20.04', 'Ubuntu-18.04', 'Fedora-36', 'Rocky-8'
```
evgeni, line 22: at which point should this be the `default` case?

evgeni (approved):

spec/classes/plugin/dns_rfc2136_spec.rb
bastelfreak: when I switch to EL9 in the next-next PR

PR #320, opened 2023-06-22
Add support for specifying key type ECDSA

martijndegouw:
Allow specifying which key type to generate. Certbot changed it's default from rsa to elliptic curve, but I kept the default to rsa for backwards compatibility.

### DIFF
--- a/REFERENCE.md
+++ b/REFERENCE.md
@@ -77,7 +77,9 @@ The following parameters are available in the `letsencrypt` class:
 * [`agree_tos`](#-letsencrypt--agree_tos)
 * [`unsafe_registration`](#-letsencrypt--unsafe_registration)
 * [`config_dir`](#-letsencrypt--config_dir)
+* [`key_type`](#-letsencrypt--key_type)
 * [`key_size`](#-letsencrypt--key_size)
+* [`elliptic_curve`](#-letsencrypt--elliptic_curve)
 * [`certificates`](#-letsencrypt--certificates)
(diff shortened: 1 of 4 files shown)


smortex:
Looks like `REFERENCE.md` is outdated.  From your working directory:

```
$ bundle update
$ bundle exec rake strings:generate:reference
$ git add REFERENCE.md
$ git commit --amend
$ git push --force
```

This will re-run the test suite.

smortex (approved):
LGTM! I'd like more feedback from module users just in case before merging.

martijndegouw:
I just quickly checked, but support for —key-type was only added in version 1.10.0 of certbot. Some of the OSses supported by this module, like Debian 10, come with an older version of certbot and will that will not work anymore with this change. 

We internally use newer versions of the certbot package for Debian 10, so I did not notice this initially. 

Should this feature take into account the actual version of certbot?

smortex:
Ah, that's the kind of tricky things I am not aware of and that can get annoying.

Maybe we can add a new fact that get the version of certbot, and does it best to respect the user config: use legacy options for old versions, maybe producing a warning if the user tuned something that cannot be tuned with their version of certbot, and use the bleeding edge options if the installed version is compatible?

kenyon:
Ubuntu 20.04 still has an old version of certbot, so we can't merge this as is. https://packages.ubuntu.com/focal-updates/certbot

bastelfreak (approved):

PR #321, opened 2023-06-22
Propagate the package_ensure parameter to all plugins

martijndegouw:
Propagate the package_ensure parameter to all plugins

### DIFF
--- a/manifests/plugin/dns_cloudflare.pp
+++ b/manifests/plugin/dns_cloudflare.pp
@@ -23,7 +23,7 @@ class letsencrypt::plugin::dns_cloudflare (
   Boolean $manage_package           = true,
   Integer $propagation_seconds      = 10,
 ) {
-  require letsencrypt::install
+  require letsencrypt
 
   if ! $api_key and ! $api_token {
     fail('No authentication method provided, please specify either api_token or api_key and api_email.')
@@ -35,7 +35,7 @@ class letsencrypt::plugin::dns_cloudflare (
     }
 
     package { $package_name:
-      ensure => installed,
+      ensure => $letsencrypt::package_ensure,
     }
   }
 
--- a/manifests/plugin/dns_route53.pp
+++ b/manifests/plugin/dns_route53.pp
@@ -16,7 +16,7 @@ class letsencrypt::plugin::dns_route53 (
 
   if $manage_package {
     package { $package_name:
-      ensure => installed,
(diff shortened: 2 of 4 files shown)


manifests/plugin/dns_route53.pp
```diff
@@ -6,17 +6,19 @@
 # @param propagation_seconds Number of seconds to wait for the DNS server to propagate the DNS-01 challenge.
 # @param manage_package Manage the plugin package.
 # @param package_name The name of the package to install when $manage_package is true.
+# @param package_ensure The value passed to `ensure` when installing the plugin package.
 #
 class letsencrypt::plugin::dns_route53 (
   String[1] $package_name,
   Integer $propagation_seconds     = 10,
   Boolean $manage_package          = true,
+  String $package_ensure           = $letsencrypt::package_ensure,
```
smortex, line 15: As long as we `require letsencrypt`, we can use `$letsencrypt::package_ensure` without the need to add parameters.  Unless there are use cases for setting an explicit different value, I would recommend not adding a new parameter.

manifests/plugin/dns_route53.pp
```diff
@@ -6,17 +6,19 @@
...
 ) {
   require letsencrypt
 
   if $manage_package {
     package { $package_name:
-      ensure => installed,
+      ensure => $package_ensure,
```
smortex, line 21: Without parameter, we can just use:

```suggestion
      ensure => $letsencrypt::package_ensure,
```

Same for the other files.

manifests/plugin/dns_cloudflare.pp
```diff
@@ -21,6 +22,7 @@
   Optional[String[1]] $email        = undef,
   Stdlib::Absolutepath $config_path = "${letsencrypt::config_dir}/dns-cloudflare.ini",
   Boolean $manage_package           = true,
+  String $package_ensure            = $letsencrypt::package_ensure,
   Integer $propagation_seconds      = 10,
 ) {
   require letsencrypt::install
```
smortex, line 28: This file require `letsencrypt::install` and not `letsencrypt`, maybe something to double-check?

smortex (commented):
Looks good, but maybe we can make this even simpler?  I am not a user of this module so may be wrong, see my in-line comments.

smortex (approved):
LGTM! I am not a user of the module so maybe another review would be welcome.

PR #322, opened 2023-06-30
Add a `certbot_version` fact

martijndegouw:
#### This Pull Request (PR) adds the certbot version as a fact.

This fact can be used to add support for new certbot features while maintaining compatibility with older version.
Note that it does not take a custom `$package_command` into account.

### DIFF
--- a/README.md
+++ b/README.md
@@ -255,6 +255,15 @@ letsencrypt::renew_deploy_hook_commands:
 
 ## Facts
 
+* [certbot_version](#fact-certbotversion)
+* [letsencrypt_directory](#fact-letsencryptdirectory)
+
+### Fact: certbot_version
+
+A fact that contains the current version of certbot installed on your operating system/distribution.
+
+### Fact: letsencrypt_directory
+
 Facts about your live certificates are available through facter. You can query the list of live certificates from puppet using `$::letsencrypt_directory` in your puppet code, hiera data or from the command line.
 
 ```
--- a/lib/facter/certbot_version.rb
+++ b/lib/facter/certbot_version.rb
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+Facter.add(:certbot_version) do
+  confine { Facter::Core::Execution.which('certbot') }
+
+  setcode do
+    output = Facter::Core::Execution.execute('certbot --version 2>/dev/null')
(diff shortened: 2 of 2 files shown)


lib/facter/certbot_version.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+Facter.add(:certbot_version) do
+  confine kernel: %w[FreeBSD Linux OpenBSD]
```
bastelfreak, line 4: we should confine on `Facter::Util::Resolution.which('certbot')` and not on the kernel I think?

martijndegouw: To be honest I copied that over from the other fact, assuming it would limit the fact to the supported operation systems only.

lib/facter/certbot_version.rb
```diff
@@ -0,0 +1,15 @@
...
+
+  setcode do
+    value = nil
+    certbot = Facter::Util::Resolution.which('certbot')
+    if certbot
+      output = Facter::Util::Resolution.exec("#{certbot} --version 2>/dev/null")
```
ekohl, line 10: `exec` is deprecated and `execute` should now be used: https://github.com/puppetlabs/facter/blob/b0d04b9aae82c1b7f58c92326667b4e2488eaf35/lib/facter/custom_facts/core/execution.rb#L90-L101

bastelfreak: ah thanks, I always mix those two up.

ekohl (approved):

smortex (commented):
Looks good, I would just reword a bit the fact description (see in-line comment: "default" vs "current").

README.md
```diff
@@ -255,6 +255,15 @@ letsencrypt::renew_deploy_hook_commands:
 
 ## Facts
 
+* [certbot_version](#fact-certbotversion)
+* [letsencrypt_directory](#fact-letsencryptdirectory)
+
+### Fact: certbot_version
+
+A Facter fact that can be used to determine what the default version of certbot is for your operating system/distribution.
```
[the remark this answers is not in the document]
martijndegouw: I updated the description

kenyon (approved):

smortex (approved):

PR #325, opened 2023-07-13
convert merge() to stdlib::merge()

slauger:
I needed one more change to make this running in my environment.

```
Error: Could not retrieve catalog from remote server: Error 500 on SERVER: Server Error: Evaluation Error: Error while evaluating a Function Call, deprecation. merge. This function is deprecated, please use stdlib::merge instead. at ["/etc/puppetlabs/code/environments/production/forge/letsencrypt/manifests/config.pp", 27]
```

We still need to fix the CI tests.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -24,7 +24,7 @@ class letsencrypt::config (
   }
 
   if $email {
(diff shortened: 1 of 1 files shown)


vchepkov:
This was fixed in #333

PR #333, opened 2023-10-16
replace obsolete merge function

vchepkov:


kenyon:
This would require stdlib >= 9, so a change to `metadata.json`.

vchepkov:
updated

manifests/config.pp
```diff
@@ -24,7 +24,7 @@
   }
 
   if $email {
-    $_config = merge($config, { 'email' => $email })
+    $_config = stdlib::merge($config, { 'email' => $email })
```
ekohl, line 27: I'd prefer
```suggestion
    $_config = $config + { 'email' => $email }
```

ekohl (changes requested):

manifests/config.pp
vchepkov: I was thinking about this exactly, then I don't need to touch metadata. Let me try one more time

ekohl (approved):

PR #332, opened 2023-10-13
plugins: avoid circular dependencies

kenyon:
`include` the letsencrypt class instead of `require` it.

Fixes #331.

PR #336, opened 2023-11-26
Add support for Fedora 39, drop support for Fedora 36

kenyon:
- plugins: avoid circular dependencies (#332)
- plugin/dns_cloudflare: require EPEL for package if needed (#332)
- metadata.json: Update supported Fedora version from 36 to 39


kenyon:
This fixes the Fedora 36 test failure in #332.

PR #337, opened 2023-11-26
move `letsencrypt::configure_epel` from hiera to `init.pp`

kenyon:
- plugins: avoid circular dependencies (#332)
- `plugin/dns_cloudflare`: require EPEL for package if needed (#332)
- `metadata.json`: Update supported Fedora version from 36 to 39 (#336)
- move `letsencrypt::configure_epel` from hiera to `init.pp`

### DIFF
--- a/data/common.yaml
+++ b/data/common.yaml
@@ -1,2 +1,0 @@
----
-letsencrypt::configure_epel: false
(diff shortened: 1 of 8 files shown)


data/common.yaml
```diff
@@ -1,2 +1,2 @@
 ---
-letsencrypt::configure_epel: false
+{}
```
bastelfreak, line 2: my personal preference is to delete the file. when I see a yam file I assume it has content

bastelfreak (approved):

data/common.yaml
kenyon: Agreed, deleted `common.yaml`.

PR #352, opened 2024-07-11
Update FreeBSD packages to Python 3.11

rtprio:
Update FreeBSD packages to Python 3.11.

This has been the default since May 29th.

On FreeBSD 14, Python versions need to be manually specified, as `py39-` packages are no longer available.

### DIFF
--- a/data/FreeBSD-family.yaml
+++ b/data/FreeBSD-family.yaml
@@ -1,7 +1,7 @@
 ---
-letsencrypt::package_name: 'py39-certbot'
+letsencrypt::package_name: 'py311-certbot'
 letsencrypt::config_dir: '/usr/local/etc/letsencrypt'
 letsencrypt::cron_owner_group: 'wheel'
-letsencrypt::plugin::dns_rfc2136::package_name: 'py39-certbot-dns-rfc2136'
-letsencrypt::plugin::dns_route53::package_name: 'py39-certbot-dns-route53'
-letsencrypt::plugin::dns_cloudflare::package_name: 'py39-certbot-dns-cloudflare'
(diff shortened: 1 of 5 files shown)


kenyon:
Maybe the FreeBSD hiera data needs to be split into files per FreeBSD major version, since this module claims to support 12, 13, and 14. Also we should probably drop 12 now that it's EOL.

kenyon (approved):
Ah I guess you would want to install the same version of certbot regardless of FreeBSD release, so this is probably good.

rtprio:
> Also we should probably drop 12 now that it's EOL.

I was going to tidy that up in a subsequent PR.

ekohl:
Does this break FreeBSD 12? In other words, is https://github.com/voxpupuli/puppet-letsencrypt/pull/354 a requirement?

smortex:
> Does this break FreeBSD 12? In other words, is #354 a requirement?

FreeBSD 12 has reached EOL last year, no new packages are build for it, so with will "break" these EOL systems.  But I think this change can be merged before the backward-incompatible changes.

When the default FreeBSD version of python is updated, all supported versions of FreeBSD get new packages targeting that new version, so the current code is currently broken on all supported FreeBSD versions and this would fix it.

Maybe a more reliable solution would be to allow to set a custom prefix (`py311`) because people can choose to use another verison of Python when they build their own packgaes.  This would make it possible for people to use the version that match their system, and would not beak when the default version of Python change or is customized.  Maybe this could be provided by the python module (where we indicate which version of python we use), but it is not a dependency of this module for now.

ekohl:
That still feels like this is a backwards incompatible change, but I'll trust you on FreeBSD specifics.

jhoblitt:
This module requires a major release anyways, so I'm adding the `backwards-incompatible` label.

PR #353, opened 2024-07-12
use a configurable root_group for things owned by root 

rtprio:
Change existing `_group` parameter to `root_group` and use it everywhere `group => 'root'` is stated.
I did not make the change backward compatible, the old parameter is renamed.

I took feedback on #231 . Since there is no `root` group, this fixes the module on FreeBSD and probably OpenBSD.


rtprio:
`letsencrypt::letsencrypt_lookup` was undocumented previously, I'm able to get that to pass on my end.

spec/classes/plugin/dns_rfc2136_spec.rb
```diff
@@ -55,7 +56,7 @@ class { 'letsencrypt':
             is_expected.to contain_file("#{pathprefix}/etc/letsencrypt/dns-rfc2136.ini").
               with_ensure('file').
               with_owner('root').
-              with_group('root').
+              with_group("#{rootgroup}").
```
bastelfreak, line 59: you can directly use the variable
```suggestion
              with_group(rootgroup).
```

manifests/scripts.pp
```diff
@@ -1,15 +1,18 @@
 # @summary Deploy helper scripts scripts
 #
 # @api private
+# @param root_group Group owner of renewal hooks and cron renew scripts.
 #
-class letsencrypt::scripts () {
+class letsencrypt::scripts (
+  String[1]                                 $root_group           = $letsencrypt::root_group,
```
bastelfreak, line 7: we don't need so much indent:
```suggestion
  String[1] $root_group = $letsencrypt::root_group,
```

PR #354, opened 2024-07-12
Remove EoL FreeBSD 12

rtprio:
Remove FreeBSD 12 from metadata and test matrix.
Add FreeBSD 14 to test matrix.

n/a

### DIFF
--- a/metadata.json
+++ b/metadata.json
@@ -70,7 +70,6 @@
     {
       "operatingsystem": "FreeBSD",
       "operatingsystemrelease": [
-        "12",
         "13",
         "14"
       ]


kenyon (approved):

TheMeier (approved):

spec/classes/plugin/dns_rfc2136_spec.rb
```diff
@@ -21,7 +21,7 @@ class { 'letsencrypt':
         case osfull
         when 'RedHat-7', 'CentOS-7'
           'python2-certbot-dns-rfc2136'
-        when 'FreeBSD-12', 'FreeBSD-13'
+        when 'FreeBSD-13', 'FreeBSD-14'
```
ekohl, line 24: If we drop EL7 support (which we now can) then this can be rewritten to simply test for the OS family and implicitly we get this for free: https://github.com/voxpupuli/puppet-letsencrypt/pull/356

Alternatively, this could simply be written like OpenBSD and remove the version specific constraint.
```suggestion
        when %r{FreeBSD}
```

Having said that, why does it pass on FreeBSD 14 both before and after this change?

PR #356, opened 2024-08-05
Drop EL7 support

ekohl:


silug:
@ekohl you'll also want to remove `data/os/CentOS/7.yaml`, correct?

I would also recommend removing CentOS 8 from `metadata.json` since it is EOL.

ekohl:
> @ekohl you'll also want to remove `data/os/CentOS/7.yaml`, correct?

I swear I looked for that. Not sure why I didn't see it.

bastelfreak (approved):
looks good to me, but can we do a minor release before merging?

bastelfreak:
https://github.com/voxpupuli/puppet-letsencrypt/pull/359

PR #358, opened 2024-09-09
Drop EoL EL7 support

bastelfreak:


bastelfreak:
ah https://github.com/voxpupuli/puppet-letsencrypt/pull/356

